PR #18, opened 2017-05-16
Change: Sized artifact wishing

NeroOneTrueKing:
If the player specifically wishes for a size of artifact (ie. large tie-dye shirt of shambhala), give it to them.


Chris-plus-alphanumericgibberish:
Hm.  Some artifacts have fixed sizes.... In light of armors now being re-sizeable (thanks for that, btw ;) ), artifact armors at least should be wishable at various sizes also.

PR #30, opened 2017-06-12
Minor fixes to curses interface

NHTangles:
Hi Chris,
FIQ found an issue in the perm_invent code if menu glyphs are enabled but menucolors are not.  This is his fix.  There may be more coming soon. I'll push them here as well if you haven't merged before then.

Cheers
Tangles.


NHTangles:
So I added another commit to this as well.  Fixes the following:
- If you start a Binder game using curses, it prompts for alignment.
- In curses, with certain display options it was not checking Binder alignment correctly, and incorrectly displaying "Lawful".

PR #37, opened 2017-09-05
Change: Advanced spells

NeroOneTrueKing:
Fireball, Cone of Cold, etc have their basic and advanced forms split.
Functional differences:
1) spells are now "Fireball" + "Fire Storm", "Cone of Cold" + "Frost Storm", "Acid Blast" + "Acid Storm", "Lightning Bolt" + "Lightning Storm"
2) no [yn] prompt on casting the advanced version; instead the player chooses from the Z menu
3) book of Infinite Spells can grant the specifically AoE versions (which are still only castable at Skilled+)


RikerW:
Cone of cold isn't really a cone anymore (it never really was), so maybe it should be renamed "Frost Blast" or similar?

RikerW:
Should the weaker, ray type spells (e.g. "Fireball", "Acid Blast") be a lower spell level than the more advanced versions? Similar to 5 vs 7, or 4 vs 6?

PR #41, opened 2017-10-04
Dagon's swamp travel, clockworks vomiting, and binders and the silver key

RikerW:
https://github.com/Chris-plus-alphanumericgibberish/dnethack/issues/24

Sorry about the cross-repo issue.

### DIFF
--- a/src/bones.c
+++ b/src/bones.c
@@ -116,6 +116,10 @@ boolean restore;
 				otmp->obj_material = COPPER;
 			    curse(otmp);
 				fix_object(otmp);
+			} else if (otmp->otyp == ART_SILVER_KEY) {
+			    otmp->otyp = KEY;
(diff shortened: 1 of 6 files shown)


RikerW:
I've also made clockworks not able to vomit.  Tarmunora suggested that any monster without innards shouldn't have to vomit, I haven't gone that far.  This only affects fountains, sinks, cursed booze, and filth (monster spell).  The only really game-changing one is the filth.

RikerW:
Also, the silver key probably shouldn't stay in a bones file.  It's kinda like an invocation artifact (though depending on role, either much easier or harder to get), and can be used instead of the bell.  

Just in case binders somehow manage to get it (despite it not occurring for them and not being in a bones file), they can no longer use it for the invocation either.

PR #46, opened 2017-11-01
bugfix: earthquake + trees interaction

NeroOneTrueKing:
Trees turned into pits wouldn't get rid of them being trees -- they'd look like pits, but you couldn't walk into them and needed an axe to chop them down!


Chris-plus-alphanumericgibberish:
Note: The reason this isn't merged is bc I think there are other bad interactions between earthquakes and terrain than just trees.

I may be confused, but it needs checking into

PR #60, opened 2017-12-22
add orcish/chaotic cave(wo)men

RikerW:
Allows caveman to be chaotic, with the addition of unrestricting orcs.


RikerW:
Closed, so later I can make an actual "orcish" caveman set-up, not just add orcs and call it good.

PR #86, opened 2018-04-19
make this work on OS X

RikerW:
This took way too many tries... With those 2 lines removed (assuming you do change the compile flag in config.h), this should compile nicely with clang on mac.


RikerW:
I'm an idiot i need to seriously make sure I pull from dev

PR #92, opened 2018-06-01
Fix '=' vs '==' mix-up

NeroOneTrueKing:
Otherwise, weapons have no safe lower limit of enchantment. Unless this was intentional?


RikerW:
This was absolutely unintentional, thanks for catching that (it was my PR).  I'm sorry, I edited that on mobile :/

PR #102, opened 2018-08-08
Reinstate Izchak's original line

AmyBSOD:
Please stop getting insulted on gay people's behalf. I'm gay (actually, lesbian) myself and I don't consider "poofy" to be insulting. Also, you should honor the late Izchak by not changing his original lines.

### DIFF
--- a/src/shk.c
+++ b/src/shk.c
@@ -4315,7 +4315,7 @@ const char *Izchak_speaks[]={
     "%s says: 'These shopping malls give me a headache.'",
     "%s says: 'Slow down.  Think clearly.'",
     "%s says: 'You need to take things one at a time.'",
-    "%s says: 'I don't like wimpy coffee... give me Columbian Supremo.'",
+    "%s says: 'I don't like poofy coffee... give me Columbian Supremo.'",
     "%s says that getting the devteam's agreement on anything is difficult.",
     "%s says that he has noticed those who serve their deity will prosper.",
     "%s says: 'Don't try to steal from me - I have friends in high places!'",


swashdev:
I concur 100% with this sentiment.  It's more offensive to presume that LGBT players can't handle a quaint word like "poofy" in a computer game.  Not all of us are shrieking narcissists whose delicate sensibilities are shredded by words.

NeroOneTrueKing:
I'm happy either way.

RikerW:
I'm also gay, and I think it should probably be kept out. I'd be in favor of just removing the line. I mean, I've never heard that slang before, but it seems to be pretty rude. For the same reason we don't call cigarettes "fags", I think we probably should keep it out.

Chris-plus-alphanumericgibberish:
My understanding of the original context of the first pull request was that it was not an "on gay people's behalf" thing but rather an actual person had been startled by that line, and requested it be changed.

In terms of what to do about it, I am rather torn.  Editing the line seemed rather Orwellian, but leaving it in seems like a disservice both to any players who **are** brought up short by it, and to Izchak himself (who, after all, presumably didn't intend to insult anyone when he said that.  Well, beyond joshing people [like me <.<] who take their coffee in the form of a hot candy bar in a cup, at least).

It might be best to just cut the line completely, or to chop it down to "Give me Columbian Supremo." or something.

swashdev:
As far as I'm concerned, the argument should be over either keeping the line as-is or deleting it. Chopping it up seems a tad disrespectful.

To me (being gay myself) this line should be preserved not for sentimental reasons but for posterity. Warner Bros. still sometimes distribute cartoons which contain blackface jokes and references to black-and-white minstrel shows, not because they're proud of those jokes but simply because they represent the era that they were made in. Censorship makes us forget both our mistakes and the people we want to remember. That's my argument, anyway.

AmyBSOD:
Blah it's not happening. Well at least I'll keep the line in dnhslex, that's for sure.

PR #106, opened 2018-10-02
Make WDSA give magical breathing

RikerW:
theoretically this could allow people to make their pets immune from stinking clouds


Chris-plus-alphanumericgibberish:
I was misremembering, I thought that WDSM DIDN'T give magical breathing to players, but it does.

PR #107, opened 2018-10-07
Update dipping for polying potions, acid

RikerW:
now, dipping a polymorphing kind of blood (chameleon, mimic), potion of polymorph will always polymorph the potion being dipped into. 
acid will explode if you dip something in it or dip it into something tha'ts not polymorph or (un)holy water.

before, dipping anything into water would dilute as normal, I think acid was wonky but polymorph definitely didn't do what it should

### DIFF
--- a/src/potion.c
+++ b/src/potion.c
@@ -2351,9 +2351,16 @@ dodip()
 		pline("That is a potion bottle, not a Klein bottle!");
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish:
Should probably also check for acidic blood  [ acidic(&mons[potion->corpsenm]) ]

PR #109, opened 2018-10-14
Refactor the pet disappearing/taming function

RikerW:
Now it's slightly easier to read, and it's standardized across files. Also fixed a small bug with god minions not being counted as "loyal" for the purposes of displaying.

### DIFF
--- a/src/allmain.c
+++ b/src/allmain.c
@@ -1126,41 +1126,13 @@ karemade:
 			}
 			/* If the player has too many pets, untame them untill that is no longer the case */
 			{
-				struct monst *curmon, *weakdog;
-				int numdogs;
-				do {
-					numdogs = 0;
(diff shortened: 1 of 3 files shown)


Chris-plus-alphanumericgibberish:
"->loyal" denotes pets that are unusually loyal to you.  They don't count against your pet cap, and they don't de-tame as a result of that cap.

Typically, the original pet is the only "loyal" pet in the game, but some roles have more on their quest home.

Chris-plus-alphanumericgibberish:
Also, this exposes a bug in the existing code: the too-many-dogs check inside the taming function *shouldn't* be exactly the same as the one in allmain: the pre-taming one is supposed to start at 1 dog rather than 0 dogs, bc it is meant to free up room for the new pet, even if the new pet is weaker than all your old pets.

PR #134, opened 2019-02-12
Fix Andro, Simurgh, Eden binder spirits

RikerW:
Andromalius - Masks should block the sign at a far distance
Simurgh - Up close, no further checks should trigger (?)
Eden - Tiny or smaller creatures should not be visible from a distance, and this should be for both far and close viewings (?)
Shiro - Move to proper alphabetical location in the logic

Thanks to carlarc for noticing Andromalius.


Chris-plus-alphanumericgibberish:
The changes to Andro have been expanded on and added, the other changes are to things that are working as intended.

The provided andromalius fix assumed that the player was wearing a blindfold, and would have crashed whenever a player w/out a blindfold but with andro bound spoke to a shk.

PR #167, opened 2019-07-02
Object materials changes and upgrades

NeroOneTrueKing:
Glyph colors are overridden in the Rogue level

The function object_color() currently only changes from the objclass color in cases of material-mismatches, but perhaps some artifacts should also have prefered colors.


NeroOneTrueKing:
This branch is still in development, but for transparency (and input!) I've set it up as a pull request. Ping me on the IRC and hopefully I'll be on to respond :)

PR #168, opened 2019-07-04
Arc redesign

RikerW:
Full artifact rundown: https://paste.ee/p/NWBZE

TL;DR: 
- Added 5 new artifacts (3 crowning gifts and 2 other artifacts currently implemented but unused).
- Redesigned the Arc quest to make the fillers not just standard room and corridor levels, along with the home and locate. 
- Merged spear and javelin skills, and added 4 new items (none of them are randomly generated, but they can be found on the aztec monsters in the quest)
 - Atlatl - a launcher for spears/javelins that uses the spear skill and crossbow rules for multishotting
 - Macuahuitl - aztec-themed club, pretty much just a club but also slashing and better damage
 - tecpatl - obsidian dagger, same damage as a droven dagger
 - ichcahuipilli - shirt that gives 3 AC
- Added 3 new monsters: Aztec warrior, spearthrower, and priest which spawn with appropriate items.

There may be something else I'm forgetting but I'm pretty sure that's it.


RikerW:
Note: I tested everything and it's all functional and all that. None of the artifacts should be particularly broken. That being said, it could use some playtesting, which I'll try to do over the next couple days. I don't expect to change anything. 

Also check out the arti placement in the Arc locate level Chris, I thought it was nice but you may want to change that. I tried to pick "ancient artifact" style artifacts that weren't particularly good/useful. That's the most objectionable thing here.

PR #169, opened 2019-07-04
make fire brand and frost brand short swords and paired

RikerW:
They are now a copper and platinum short sword respectively, which can be two-weaponed with each other only. (to use them effectively you must be skilled, they're both set to 30 wt (standard short sword)) Invoking one before the other is generated will create it ala blade singer/dancer.

I'm not expecting this to be pulled but I figured I'd make the PR anyway just so you could see what I did on my local copy. 

Fun fact: fixed a longstanding 'bug' where there was no punctuation "An object appears at your feet" when you summoned the other bladesinger pair.


NeroOneTrueKing:
Cool!
I think there was a YANI a while back, something like:
"make Frost Brand and Fire Brand twoweaponable, made of mithril, and have a randomized-each-game shared base type (some kind of light one-handed weapons)"
😄

RikerW:
whoops I did not mean to merge my arc changes into here. I'll probably fix that later.

Chris-plus-alphanumericgibberish:
I took these in a slightly different direction (enhancing the fiery/frosty damage and letting them still do +.5x damage to resistant targets a la Elemental attacks).

I did pull the "can be two-weaponed together" angle, but didn't link them via #invoke

PR #196, opened 2019-07-24
Allow mode-switches on the '+' spell menu when only knowing 1 spell

NeroOneTrueKing:
Thanks for the bug report, LarienTelrunya


AmyBSOD:
:smiley: Thanks!

PR #205, opened 2019-07-27
Shadowsteel no longer evaporates while invisible, more Spidersilk changes

RikerW:
This protects shadowsteel worn or carried from being afflicted by light if the bearer (you or a monster) is invisible, however fixed shadowsteel will not regenerate while you're invisible and in darkness (or just plain invisible). The justification for this that it needs to be submerged in shadow to heal.

Also, Spidersilk now grants an extra 2 AC/DR to compensate for the lack of protection that cloth gives, and shows as being made of "spidersilk" instead of "cloth" in the more info menu as well as when unidentified ("Made of spidersilk.", "a spidersilk droven chain mail named Spidersilk").

### DIFF
--- a/src/do_wear.c
+++ b/src/do_wear.c
@@ -2014,6 +2014,9 @@ struct obj * otmp;
 	case ART_AEGIS:
 		def += 3;//3 total, Same as Cloak of protection
 		break;
+	case ART_SPIDERSILK:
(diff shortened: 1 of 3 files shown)


Chris-plus-alphanumericgibberish:
Merge pending in next batch

PR #211, opened 2019-08-06
Fix items having wrong hands when held by monsters

NeroOneTrueKing:
In reference to https://github.com/Chris-plus-alphanumericgibberish/dNAO/projects/3#card-24372042


NeroOneTrueKing:
The check failed, but I think that was a server connection issue and not a code issue:
https://travis-ci.org/Chris-plus-alphanumericgibberish/dNAO/jobs/568548604

PR #216, opened 2019-08-17
For some reason, this is necessary.

NeroOneTrueKing:
Testing concluded the following:
0x00000000ffffffff == ~(0xffffffff00000000)
0xffffffff00000000 != ~(0x00000000ffffffff)

So let's not use '~' with long long integers.


NeroOneTrueKing:
Chris pointed out this is not the correct solution.
These should all be long literals, with an 'L' at the end.

PR #242, opened 2019-09-18
Make arcs actually receive their crowning gifts

RikerW:


Chris-plus-alphanumericgibberish:
Send bug fixes to compat and new features to devel, devel merges from compat.

PR #259, opened 2019-10-07
Let you aim Ruinous Descent of Star's starfall 

RikerW:
Otherwise it's kinda hit or miss whether it's useful or not


RikerW:
ah hell I can't use github

PR #266, opened 2019-11-06
block drow from getting atma weapon as a gift

RikerW:
they already can't do the arkenstone or holy moonlight sword, now atma weapon too. this is because it just blinds them unless they're willing to wish (?) for fossil dark, since I've never seen it in-game beside from a wish


Chris-plus-alphanumericgibberish:
The lack of fossil dark is the bug.  It's meant to generate but there was an error with integer math giving it 0 probability.  There is also a non-magic gem that would work, if any happen to spawn.

PR #271, opened 2019-11-11
Add relevant skills for (he)drow shared

RikerW:
Now, all female drow shared should be able to get expert in saber (original QA, traitor alt), flail (traitor), crossbow (original alt). In addition, hedrow can get expert in polearms and flail (traitor artifacts). 

I did NOT give hedrow any skills for Spidersilk, but I guess they could get martial arts when given spidersilk maybe? and possibly skill in attack spells from the start, not sure. Darkweaver's doesn't seem to have any direct associations (and it's not an amazing artifact either in my humble opinion), so it gets nothing.

### DIFF
--- a/src/u_init.c
+++ b/src/u_init.c
@@ -1005,12 +1005,6 @@ static const struct def_skill Skill_Drow_Unarmed[] = {
     { P_NONE, 0 }
 };
 
-static const struct def_skill Skill_DP[] = {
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish:
Not all roles should be able to use all quest artifacts equally well.

PR #341, opened 2020-01-10
Link statue traps to statues

### LINKED ISSUE #331: Stoned monsters bug statue traps
Stoning a monster that has managed to move onto a statue trap square without triggering the trap replaces the old statue with the new statue corpse, "disarming" the trap.

NeroOneTrueKing:
First, to their generated-with statue.

If statues are placed over top of them in sp_lev, re-link to the new statues

Fixes #331

### DIFF
--- a/include/trap.h
+++ b/include/trap.h
@@ -17,9 +17,14 @@ struct trap_loaded_obj {
 };
 
 union vlaunchinfo {
+	/* 16 bits */
+	/* arrow/dart */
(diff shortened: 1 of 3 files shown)


Chris-plus-alphanumericgibberish (approved):
obfree looks like it was called too soon.  I think you'd need to call at minimum obj_extract_self before calling obfree, otherwise it will break the list.

PR #347, opened 2020-01-11
Create vampiric minions

RikerW:
Vampires can create tame minions from a drained corpse. The corpse must be less than 20 turns old (limit designed that for something like a dragon, you have just enough time to kill & drain it before it ages too much). 

Limited by:
- must have a drained corpse on the floor or in your inventory
- the monster must have blood (should be redundant with drained anyway)
- the monster must be non-unique and tameable

Creates a new monster of that type with the vampiric faction added, uses up the corpse, and drains a level. Accessible to all vampires level 2+ (including polyselfed).


RikerW:
The Staff of Necromancy:
* vampire favoring quarterstaff
* +1d5 to-hit, double life-draining damage
* grants cold resistance and curse resistance when wielded
* invoke to resurrect one corpse (of a monster with bones) as a tame skelefied version 

TODO: make it grant a bonus to drain life / turn undead spells? unsure, will think about

Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish (approved):

PR #348, opened 2020-01-13
Add callandor

RikerW:
Callandor

* artifact crystal sword
* when wielded by a male character
   * grants energy regeneration, half spell damage, and boosts spells while wielded
    * on 1/20 of spell casts, subtracts rnd(spell_level) sanity

Reference to Callandor, a male sa'angreal from the Wheel of Time series. Notable that it amplified the taint, thus the the san loss, but also heavily boosted spells. Possibly could have a negative effect of some sort on your health as well after prolonged usage (ala tensa z?). 

I'm not really expecting this to be merged, just fyi. If you do that's cool, but it probably needs some more fine tuning.

### DIFF
--- a/include/artilist.h
+++ b/include/artilist.h
@@ -595,6 +595,14 @@ A("Fuma-itto no Ken",		BROADSWORD,		0,			0, /*Needs encyc entry*/
  *  Nethack Samurai call broadswords "Ninja-to," which is the steriotypical ninja sword.
  *  Aparently, there was no such thing as an actual Ninja-to, it's something Hollywood made up!
  */
+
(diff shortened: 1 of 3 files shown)


NeroOneTrueKing:
Misc suggestions:
Auto-curse when wielded?
Use current (rather than initial) gender?

RikerW:
It is intentional that it uses initgend. There's a character in the books (aginor?) reincarnated as a female, but that still touches the male power. Ideally I'd also case this to only be given to initgend males, and possibly add a female one as well later at some point (there aren't as many notable female angreals).

Autocursing is a thought, but my goal is to try to make it easily usable with but with a price (which could get annoying with cursing). It may thematically fit though.

Chris-plus-alphanumericgibberish (approved):

PR #386, opened 2020-01-26
Patch room weirdness

### LINKED ISSUE #385: Glyph for a doorway can sometimes be funky in conjoined rooms
https://paste.ee/p/ToUG9

The doorway (hard to see, but the bottom room's top wall) opens left to right instead of up and down (like it was a horiz doorway instead of a vertical one). It opens and closes fine, and appear as a plus when closed as expected. Movement through it is vertical, despite the glyph.

NeroOneTrueKing:
Fixes #385

### DIFF
--- a/src/mklev.c
+++ b/src/mklev.c
@@ -489,25 +489,22 @@ merge_adj_rooms()
 				smeq[j] = smeq[i];
 			else
 				smeq[i] = smeq[j];
-			// make the lighting consistent between the rooms -- this will fail sometimes when merging already-merged rooms, but this isn't critical
-			if (a->rlit != b->rlit)
-			{
-				int lit = 0;
-				struct rm *lev;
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (commented):
The fact that the horizontal bit is only set in 1 of the 4 room expansion loops seems surprising to me.  Should it also be set in one of the other three?

NeroOneTrueKing:
1/4 seems weird at first, but it is correct (I think).
When we're expanding the room by placing ROOM tiles, we want to set horizontal to 0. That's 2/4 cases.
When we're expanding by placing VWALL and HWALL, the former is vertical. That's 1/4, which leaves only the placing of HWALL to set horizontal=1.

Chris-plus-alphanumericgibberish (approved):

PR #389, opened 2020-01-26
Magic chests bolt themselves to the floor when first unlocked

NeroOneTrueKing:
This way, you can choose to move a magic chest to your stash (or wherever it is you want it) before placing it, but you cannot carry one around with you as a chest of holding.

Once bolted down, they are very hard (impossible, hopefully) to move.
* Attempting to pick one up gives a message that the chest is bolted down.
* Being magically fixed in place, a bolted magic chest cannot be teleported away or kicked around.
* Gelatinous cubes don't engulf them, and other monsters won't pick them up.
* The chest won't fall from other items impacting it
* The chest is also set to weigh a ridiculous amount while bolted down.

The Master Key of Thievery's untrapping invoke is the only way to unbolt one.

Magic chests also cannot be opened by kicking or forcing the lock.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -2616,7 +2616,7 @@ E boolean NDECL(drown);
 E int NDECL(dodeepswim);
 E void FDECL(drain_en, (int));
 E int NDECL(dountrap);
-E int FDECL(untrap, (BOOLEAN_P));
+E int FDECL(untrap, (struct obj *));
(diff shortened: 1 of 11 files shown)


RikerW (dismissed):
Nothing to see here.

Chris-plus-alphanumericgibberish (approved):

PR #393, opened 2020-02-01
Make acid destroy iron bars when thrown

RikerW:
I believe unnethack adds this for force-fighting, here it is for throwing.

Not a bug, but a feature of sorts. For best results use with nero's venom fix.


RikerW:
Pretty sure this won't compile until Nero's fix is merged, so I'm closing.

PR #408, opened 2020-02-12
Change Astaroth's pen binding

RikerW:
Instead of requiring the pen to be rusty, there needs to be a rusty/etc. item that's a weapon/weptool and is skill P_DAGGER.

For choosing the item, instead of choosing the first rusty/etc. item in the pile it will choose the last, but break early if the selected item is P_DAGGER (so using the first one of those it finds)

I'm not sure this is an appropriate change, but I'm putting it out there. The other solution(s) are things like making the pen not rustproofed when given by your god, which I'm personally not a fan of, but could work.


Chris-plus-alphanumericgibberish:
The Pen binding rituals should all focus on the Pen where possible.  I have removed errodeproofing from a gifted Pen to allow it to be used in this ritual.

PR #418, opened 2020-02-22
Always reset killer after hmon

### LINKED ISSUE #339: Prismatic Spray can lead to printing weird strings
Monster spell prismatic spray, when it kills a monster, will print that the monster is caught in the %s but the thing in the %s tends to be inventory items or other previously recently seen strings.

NeroOneTrueKing:
Hopefully fixes #339.

### DIFF
--- a/src/xhity.c
+++ b/src/xhity.c
@@ -205,20 +205,20 @@ struct monst * mdef;
 		weave_black_web(mdef);
 	
 	/* Do the attacks */
-	bhitpos.x = u.ux + u.dx; bhitpos.y = u.uy + u.dy;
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (changes requested):
It looks like this was against an earlier version of xhity, and is regressing a bunch of stuff.

Examples:
magr_can_attack_mdef(), 
Clear_thoughts,
and some bugfixes involving what the HP thresholds for dying actually are.

NeroOneTrueKing:
Whoa, that is weird.
I remember what the change was, so I'll close this and re-create it from current devel-3.19.0

PR #421, opened 2020-02-23
Improve hmon, particularly with traps

NeroOneTrueKing:
Key features:

Hmon can now distinguish between many things (melee, thrust polearm, projectile, misthrown projectile, trap), and can easily be expanded to cover more.

Most traps were made to use this. Falling rock traps in particular are now a basic 1d6-[normal DR and whatever], instead of 2d6 [max 2 if wearing hard helmet]. Dart and arrow traps now use hmon vs monsters (oops that I hadn't done that yet).

Fixed a bug with having a launcher in your swapwep.

Arrow/dart traps deal appropriate damage -- I noticed that they were doing wimpy d4-ish damage before, which means just about nothing after XL1.


NeroOneTrueKing:
Oh yeah, I was going to make beartraps "hit" the creature they close on to deal searing damage and stuff. Also there'll surely be other things that need fixing. Riker, if you feel like pulling this branch and play-testing it and telling me what you think, awesome. I don't think it breaks saves vs devel-3.19.0.

Chris-plus-alphanumericgibberish:
o.O what happened with trap.c?  Was that because I resolved a merge conflict via the web interface?

...Looks like it >_<

Chris-plus-alphanumericgibberish (dismissed):

Chris-plus-alphanumericgibberish (approved):

PR #422, opened 2020-02-23
Extracting an object from a chain resets obj->nobj

NeroOneTrueKing:
Do not require caller of extract_nobj() to (re)set obj->nobj.

### DIFF
--- a/src/mkobj.c
+++ b/src/mkobj.c
@@ -2865,6 +2865,7 @@ extract_nobj(obj, head_ptr)
 		panic("extract_nobj: object lost");
 	}
     obj->where = OBJ_FREE;
+	obj->nobj = (struct obj *)0;
 }
 
 


Chris-plus-alphanumericgibberish (commented):
Have you checked over all the places this gets called to make sure nothing is assuming that nobj is still pointing at the rest of the chain?

replace_object() is doing *something* really freaky with nobj, but I think it's faking the removed object still being on the chain.

NeroOneTrueKing:
I hadn't, but now I have.

`replace_object()` is manually inserting `otmp` into the object chain behind `obj`, which means that `obj->nobj` has to point to `otmp`... so that `extract_nobj()` (which is called directly afterwards) is working with a valid chain.

You'd need to be doing really klunky stuff to not want `nobj` cleared -- you'd expect `extract_nobj()` to completely separate the object you pass it from whatever chain you think it's in.

Chris-plus-alphanumericgibberish (approved):

PR #433, opened 2020-03-01
Blasphemous Lurkers should not be wishable.

NeroOneTrueKing:
Should they even be tameable?


Chris-plus-alphanumericgibberish (approved):
Re: tameable: No. Already fixed on local.

PR #473, opened 2020-04-07
Blade singer/dancer pair are now a rakuyo

RikerW:
the singer is the "rakuyo", which turns into the rakuyo-saber when it's unlatched
the dancer is the "rakuyo-dagger", which is first made when the rakuyo is first unlatched (it still technically exists even when the rakuyo is latched and the item is gone for artiexist purposes)

when an ARTA_HASTE weapon is hit with, it adds normal speed/3 pts to your movement. if your swapwep is an artifact with ARTA_HASTE, it also adds speed/6 pts to your movement. that does NOT require you be twoweaponing, since i can't see that being abused easily and it's a nice bonus if you need to use a shield for a bit.

they're still both silver, which is pretty good. the damage dice were not majorly changed, just moving +6/6 to +8/8 on the saber to represent the slightly better damage dice

you will never be given the blade dancer's dagger, only the blade singer's *saber* (it's been renamed now, for obvious reasons). likewise, the dagger is not only marked as NOGEN but shouldn't even have a chance to spawn assuming nobody starts placing rakuyo-daggers by themselves or unlatched rakuyo bits

lastly, the invokes were changed. previously they used to spawn the other one (including: worthless chaff if the mate already existed, punctuation errors). the arti invoke has been renamed from SING_SPEAR (spawned spear) / DANCE_DAGGER (spawned dagger) to BLADESONG

the bladesong:
- heals up to 1/2 your lost hp (`(max - current)/2`), scaled by the spe of the weapon up to the full 1/2 at +10, a total of 1/4 healing at +5, etc.
- temporarily gives you very fast speed (ala haste self / speed monster)
- gives you protection equal to lvl/10 + spe, lasting for `(lvl * (50|30|10))/15` turns (based off BUC)

if the spear is being invoked with the dagger offhanded, the speed bonus duration and the protection is doubled. the health is not. again, no two-weaponing required, but it does take your swapwep. the dagger being invoked will never have the double bonus, it's supposed to always be strictly better to invoke the spear

this has not been balance-tested at all whatsoever. in addition, since the insight changes haven't been pushed, we'll see how good this turns out to be (i'm expecting a nerf). in total, it's a much better weapon, but that doesn't mean a ton considering it was kinda bad before. this is supposed to be competitive with a two-handed artifact weapon and strictly better than two-weaponing a normal artifact + bane (or normal + any artifact that doesn't object to being offhanded).


RikerW:
Somehow this lost the do_rakuyo changes, I'll fix that.

Chris-plus-alphanumericgibberish (approved):

PR #479, opened 2020-04-10
fix most of the warnings, and ignore any k&r ones

RikerW:
warnings were mostly k&r ones (which I ignored, since there's not much I can do, and that would require rewriting a lot of useless lines)

the ones i fixed classified under "need parens for assignment in if/while statements", "|| wthin &&" and some "== when you mean =".

For reference, here are some fun files. The stderr output pre-commit, the stderr output post-commit, and a modified diff (I edited out the "-Wno-knr" from the file for the diff so it didn't spam every line).

[makeoutput.txt](https://github.com/Chris-plus-alphanumericgibberish/dNAO/files/4459332/makeoutput.txt)
[makeoutput.txt](https://github.com/Chris-plus-alphanumericgibberish/dNAO/files/4459333/makeoutput.txt)

[warnings.txt](https://github.com/Chris-plus-alphanumericgibberish/dNAO/files/4459336/warnings.txt)

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -3328,7 +3328,7 @@ resistances_enlightenment()
 */
 	if(u.umconf){
 		if(Blind) Sprintf(buf, "Your %s are tingling.", makeplural(body_part(HAND)));
-		else Sprintf(buf, "Your %s are%s glowing%s red.", makeplural(body_part(HAND)), u.umconf > 20 ? " brilliantly" : u.umconf > 10 ? " brightly" : "", hcolor(NH_RED));
+		else Sprintf(buf, "Your %s are glowing%s %s.", makeplural(body_part(HAND)), u.umconf > 20 ? " brilliantly" : u.umconf > 10 ? " brightly" : "", hcolor(NH_RED));
 		putstr(en_win, 0, buf);
 	}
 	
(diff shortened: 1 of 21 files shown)


src/cmd.c
```diff
@@ -3328,7 +3328,7 @@ resistances_enlightenment()
 */
 	if(u.umconf){
 		if(Blind) Sprintf(buf, "Your %s are tingling.", makeplural(body_part(HAND)));
-		else Sprintf(buf, "Your %s are glowing%s red.", makeplural(body_part(HAND)), u.umconf > 20 ? " brilliantly" : u.umconf > 10 ? " bright" : "", hcolor(NH_RED));
+		else Sprintf(buf, "Your %s are%s glowing%s red.", makeplural(body_part(HAND)), u.umconf > 20 ? " brilliantly" : u.umconf > 10 ? " brightly" : "", hcolor(NH_RED));
```
Chris-plus-alphanumericgibberish: "are%s" is wrong, "red" should be replaced with "%s"

Chris-plus-alphanumericgibberish (changes requested):
Could you fix the cmd.c problem?

Chris-plus-alphanumericgibberish (approved):

PR #482, opened 2020-04-11
Nerf vorpal weapons [vs bosses]

### LINKED ISSUE #444: Vorpal things should not decapitate monsters past a certain hp threshold
Vorpal Blade (and other vorpal weapons) are broken from a gameplay point of view - they can instantly kill any non-headless monster in the game with no save once you get past its AC. 

A solution to this would be to make vorpal weapons only "decapitate" (read: instakill) things under a certain hp threshold. That way, you can still use the instakill effect, without making it so busted on uniques (especially the tankier ones, like Asmo and {Demo}).

NeroOneTrueKing:
Bonus damage = 12X base + 3d20, instead of instakill.

Bonus damage can be reduced by relevant armor (helmets vs vorpal blade, etc), which damages and destroys armor. If the armor isn't strong enough, the vorpal attack can still be lethal after destroying the armor.
Armor does not have an effect if the vorpal weapon is arti_shining (Shadowlock).

Exact numbers may need tweaking (how much bonus damage? how effective is armor?).

Closes #444

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -3877,7 +3877,7 @@ boolean * messaged;
 	/* vorpal weapons */
 	if (arti_attack_prop(otmp, ARTA_VORPAL) || (oproperties&OPROP_VORPW)) {
 		char buf[BUFSZ];
-		int vorpaldamage = (basedmg * 12) + d(3, 20);
+		int vorpaldamage = (basedmg * 12) + d(5, 20);
 		int method = 0;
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing:
Potential tweak:
`vorpaldamage = max(20*basedmg, 300)`
Thoughts?

Chris-plus-alphanumericgibberish (approved):

PR #500, opened 2020-04-15
Give the Marauder's Map an appearance

NeroOneTrueKing:
This will replace "a scroll labeled READ ME named the Marauder's Map" with "a parchment scroll".


NeroOneTrueKing:
From lore:
> The map is normally disguised as a blank piece of parchment. To view the map, one must tap it with one's wand and recite, "I solemnly swear that I am up to no good." The content of the map will then reveal itself.

Hmm, possibilities...

Chris-plus-alphanumericgibberish (approved):

PR #503, opened 2020-04-15
Bugfixes: your zaps, monster zaps

NeroOneTrueKing:
Your zaps start at u.ux, u.uy.
Bug created by #454


NeroOneTrueKing:
Added two more zap fixes, relating to monster usage of wands (and other offensive items)

Chris-plus-alphanumericgibberish (approved):

PR #531, opened 2020-04-26
 cavemen get +1 size, ymir grants +1 size

RikerW:
this applies for weapon wielding only - both the handedness (bimanual in wield.c) and the to-hit modifier in xhity/invent.c

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -2863,6 +2863,8 @@ winid *datawin;
 				size_penalty = max(0, size_penalty - 1);
 			if (u.sealsActive&SEAL_YMIR)
 				size_penalty = max(0, size_penalty - 1);
+			
+			hitbon -= size_penalty * 4;
 		} 
 		
 		if (hitbon != 0)
--- a/src/invent.c
+++ b/src/invent.c
@@ -2855,7 +2855,16 @@ winid *datawin;
 		}
 
 		/* to-hit */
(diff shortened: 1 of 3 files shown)


src/invent.c
```diff
@@ -2855,7 +2855,16 @@ winid *datawin;
 		}
 
 		/* to-hit */
-		int hitbon = oc.oc_hitbon - (obj ? (4 * max(0,(obj->objsize - youracedata->msize))) : 0);
```
Chris-plus-alphanumericgibberish, line 2858: I think that 4x disappeared?

src/xhity.c
```diff
@@ -3293,9 +3293,16 @@ int flat_acc;
 			/* Houchou always hits when thrown */
 			if (weapon->oartifact == ART_HOUCHOU && thrown)
 				wepn_acc += 1000;
+
 			/* -4 accuracy per weapon size too large (not for thrown objects) */
-			if (!thrown && weapon->objsize - pa->msize > 0){
-				wepn_acc += -4 * (weapon->objsize - pa->msize);
```
Chris-plus-alphanumericgibberish, line 3298: A 4x also seems to have gone missing here.  The behavior is changed, too, small objects used to get a bonus.  However, that seems wrong and the comment agrees.

Chris-plus-alphanumericgibberish (approved):

PR #536, opened 2020-04-27
Change title_to_mon() to only match exact length titles

### LINKED ISSUE #519: Genociding "master mindflayer" as a monk genocides you
C343-395
https://github.com/NetHack/NetHack/commit/fe5d79c85943314831f6a0a931aece8401462194

NeroOneTrueKing:
Prevents "master mindflayer" matching "master", for example. (fixes #519)


NeroOneTrueKing:
Ugh, it's not this easy, because wishing code needs to be able to give it something like "master mask".

PR #567, opened 2020-05-09
Improvements to 3d water

NeroOneTrueKing:


src/trap.c
```diff
@@ -3528,23 +3528,25 @@ drown()
...
+		} else if(is_3dwater(u.ux,u.uy)){
 			u.usubwater = 1;
 		} else if(Amphibious && !Swimming){
 			u.usubwater = 1;
 		}
 		if (Amphibious &&  u.usubwater){
 			if (flags.verbose)
 				if(!Swimming) pline("But you aren't drowning.");
-			if (!Is_waterlevel(&u.uz)) {
+			if (!is_3dwater(u.ux, u.uy)) {
```
Chris-plus-alphanumericgibberish: This distinction was actually purposeful, the Plane of Water is an endless ocean with no top or bottom, while Dagon's level has you walking around on the seafloor.

Chris-plus-alphanumericgibberish (approved):

PR #569, opened 2020-05-09
Patch buff xiuhcoatl

RikerW:
It's "serpenthide" (still dragonhide but a new description) now, and it deals a bonus 2d4 damage. The 2d4 is from the couatl bite attack damage. No poison, since that would suck if you're lawful.

Also fix a minor bug - Y'ha-Talla and Xiuhcoatl both say "What is your command, my Lord" when invoked (since it uses the LORDLY invoke for toggling forms), and they'd refer to you as a "Lord" regardless of gender. Now only the 3 actual lordly weapons (lordly might, elvish lords, sceptre of lolth) give you a title, and they'll gender you correctly.

### DIFF
--- a/src/objnam.c
+++ b/src/objnam.c
@@ -1000,6 +1000,8 @@ boolean adjective;
 		/* for some reason, this is dragonhide? */
 		if (obj->otyp == LEO_NEMAEUS_HIDE)
 			return "lionhide";
+		if (obj->oartifact && obj->oartifact == ART_XIUHCOATL)
+			return "serpenthide";
 		/* hard object made of dragonhide (or described as being dragon-bone) -> bone or tooth */
 		else if ((objects[obj->otyp].oc_material > LEATHER && objects[obj->otyp].oc_material != DRAGON_HIDE)
 			|| ((s = OBJ_DESCR(objects[obj->otyp])) != (char *)0 && !strncmp(s, "dragonbone", 10))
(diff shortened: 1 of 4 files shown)


src/objnam.c
```diff
@@ -1000,6 +1000,8 @@ boolean adjective;
 		/* for some reason, this is dragonhide? */
 		if (obj->otyp == LEO_NEMAEUS_HIDE)
 			return "lionhide";
+		if (obj->oartifact && obj->oartifact == ART_XIUHCOATL)
```
NeroOneTrueKing: Asking if `obj->var == const` is equally safe as asking if `obj->var` exists -- both will segfault if `obj` does not exist. That is to say, in general, the check you'd need to use is `if(obj && obj->oartifact == const)`.

In this particular case, we are already assuming that `obj` exists, so no additional checks are necessary, and we can just use `if(obj->oartifact == ART_XIUHCOATL)` directly.

src/weapon.c
```diff
@@ -513,6 +513,14 @@ int otyp;
 			bonn = 1;
 			bond = 4;
 		}
+		else if (obj->oartifact == ART_XIUHCOATL && otyp == BULLWHIP)
+		{
+			// couatl bite 
+			ocn = 1;
+			ocd = 2;
+			bonn = 2;
+			bond = max(4 + 2 * dmod, 2);
```
NeroOneTrueKing: As this is meant to be equal to the bite of a Couatl, perhaps it should follow the same logic as the Golden Sword of Y'ha-Talla and not be affected by dmod. Thoughts?

NeroOneTrueKing (commented):
Looks good to me, just me being a nitpick 😄

Chris-plus-alphanumericgibberish (approved):

PR #571, opened 2020-05-10
Split ^V leveltele menu into 2 inputs

NeroOneTrueKing:
So you can actually levelteleport via menu to somewhere that isn't the main dungeon or gehennom.

There appears to be further issues related to the teleportation itself that *hopefully* this addition has only exposed, not created. Teleporting to Vlad's Tower via the menu is spotty, for one.


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
...I haven't had problems with levelporting via the menu?

PR #593, opened 2020-05-19
Adjust Free Action + Stoning interaction

NeroOneTrueKing:
Free Action now gives you 3 more turns to take actions when afflicted by delayed stoning.
However, free action no longer protects against the "Turn to Stone" monster spell -- better carry lizard corpses!


RikerW:
Clarification for future readers: with free action, you can now move during the final 3 turns of being stoned (normally, only the first 2 can be moved on and the rest are stunlocked, so a 5 turn timer is really 2). this does NOT increase the overall timer to 8 or anything like that.

Chris-plus-alphanumericgibberish (approved):

PR #632, opened 2020-05-26
Axus' auton spawns get no inventory

NeroOneTrueKing:


NeroOneTrueKing:
This is only a partial fix:  his auton spawns are still worth score and XP.

Chris-plus-alphanumericgibberish (approved):

PR #655, opened 2020-06-01
Reading the Blinding Glyph should grant Intrinsic Flight

NeroOneTrueKing:
Fixes bad messages involving flying boots after having read the Slab.

The `u.ufirst_light` check in youprop.h should be removed with the next save-breaking patch.

Also, replace bad mon_resistance(&youmonst, ...) checks.

### DIFF
--- a/src/do_wear.c
+++ b/src/do_wear.c
@@ -135,7 +135,7 @@ Boots_on()
 			incr_itimeout(&HFumbling, rnd(20));
 		break;
 	case FLYING_BOOTS:
-		if (!oldprop && !(mon_resistance(&youmonst,FLYING) || (u.usteed && mon_resistance(u.usteed,FLYING))) && !HLevitation) {
+		if (!oldprop && !(species_flies(youracedata) || (u.usteed && mon_resistance(u.usteed,FLYING))) && !HLevitation) {
 			makeknown(uarmf->otyp);
 			float_up();
 			spoteffects(FALSE);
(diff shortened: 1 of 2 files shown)


Chris-plus-alphanumericgibberish (dismissed):
This overloads the FROMOUTSIDE bit (which is already overloaded >_<).  In general, things that do not use that bit currently should continue to not use that bit. It's possible that separate bits could be dedicated to each additional source of an intrinsic.

Unrelated note: specifying things and an INTRINSIC in the prop.h-constant-sense should also be done very carefully, since for some intrinsics it grants gremlins the ability to steal the intrinsic.

NeroOneTrueKing:
From my understanding, `FROMOUTSIDE` is already a catch-all for permanent intrinsics. Thrones, crowning, unique corpses, eating jewelry, etc. With the sole exception of `AD_CURS` attacks (gremlins, Glasya, Phantasms, and the Avatar of Lolth), they cannot be revoked. If the property isn't one of the 11 stealable ones, it will remain for the rest of the game and overloading it is safe.

I think that, if anything, we need to specify a particular bit in `prop.intrinsic` that gremlins can steal from. Some properties would be permanent-until-stolen, and all others would be permanent-forever. The latter can then be safely overloaded by any number of sources, because by design it will never be revoked.

Chris-plus-alphanumericgibberish:
Ok, yeah, defining FROMOUTSIDE  as a catch-all bit for permanently gaining the intrinsic would work.

...If anything, gremlins should probably be able to steal timed intrinsics, and not permanent intrinsics of any type.

Chris-plus-alphanumericgibberish (approved):

PR #703, opened 2020-06-09
Elf crowning gifts had bad otyps

RikerW:


NeroOneTrueKing:
Uh, rather, I think I mixed them up in artilist.h when I changed that file

NeroOneTrueKing (changes requested):
Go to artilist.h and fix the otyps that I messed up there, instead. This code was correct.

NeroOneTrueKing (approved):
OOOps, and I see there was a commit by Chris swapping the otyps around, e9a301acb1cc9d23a4640a91771d03ab7098c413.

PR #715, opened 2020-06-12
Fix impossible() in library of ill-regard monster placement

### LINKED ISSUE #674: Library of Ill Regard doesn't like kraken
Now that makemon() checks "hey can we actually put monster here" by default, it refuses to place a kraken onto dry land.

Possible solutions:
 - don't attempt to place sea creatures in Ill-Regard
 - change goodpos to make vivisection traps acceptable for all creatures (and move trap creation before monster creation at ln2332 in mkmaze.c)
 - have a makemon flag to say "don't check goodpos" <- not recommended

NeroOneTrueKing:
Kraken (all `;`) are not eligible.
Fixes #674

### DIFF
--- a/src/mkmaze.c
+++ b/src/mkmaze.c
@@ -2297,7 +2297,7 @@ fill_dungeon_of_ill_regard(){
 		}
 	}
 	for(i = 0; i < PM_LONG_WORM_TAIL ; i++){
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (approved):
This is technically resolving the problem in the wrong direction, but clearing the impossible is the higher priority.

PR #745, opened 2020-06-14
Fix Ahazu abductions crashing on riders

NeroOneTrueKing:
Riders shouldn't be able to be disposed of this way, either.


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
It actually was intended that they be a possible target, but clearly it was never tested.

I'm also not really sure that there *should* be a way to get rid of them :-\

PR #792, opened 2020-06-20
Don't add extra cap to yuki-onna's str/con carry amount

### LINKED ISSUE #785: Yuki carrycap is sometimes less than it should be
Nymphs have their carrycap set to MAX_CARRY_CAP in hack.c. However, MAX_CARRY_CAP is 1000, but a 25 str/con character has 1300 already. When you add in other bonuses like hiking boots, it's pretty easy to have a carrycap that's too low. This is mostly seen with yuki, who you're likely to wish ogresmasher on.

NeroOneTrueKing:
Nymphs are supposed to be able to carry lots, so they get the default MAX_CARRY_CAP, 1000.
When the player has +maxcarrycap, they can get a maxcarrycap > 1000.
Don't limit the player's carrycap to 1000 prematurely.

Fixes #785

### DIFF
--- a/src/hack.c
+++ b/src/hack.c
@@ -2650,7 +2650,7 @@ weight_cap()
 	}
 	/* consistent with can_carry() in mon.c */
 	if (mdat->mlet == S_NYMPH)
-		carrcap = MAX_CARR_CAP;
+		carrcap = max(carrcap, MAX_CARR_CAP);
 	else if (!mdat->cwt)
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (approved):
This seems like it won't change the calculation any for yukis, since their max capacity is lower than the default max capacity (though it would affect a giant nymph)

PR #799, opened 2020-06-21
Fix dragon breaths being easily reflectable

NeroOneTrueKing:


Chris-plus-alphanumericgibberish:
Considering that we've gone through most of June with this in place, lets not fix it until after June.

Chris-plus-alphanumericgibberish (approved):

PR #802, opened 2020-06-21
Close loopholes for Anas to lose telepathy

NeroOneTrueKing:
Being not-telepathic makes their quest significantly easier. Anachrononauts were changed to start with telepathy for a reason.
Therefore, remove methods for losing telepathy for anachrononauts.


NeroOneTrueKing:
Update: was prompted to double-check if all ana really do get telepathy at start; it's just incantifier and elven anachrononauts that do.

Maybe this is not necessary.

Chris-plus-alphanumericgibberish:
Yeah, delaying getting telepathy was intended to be an option, and removing it via murder is an option if you get it too soon.

Possibly telepathy gained as a result of race or role shouldn't be removable, but their needs to be a lead-lined helmet or something that can block telepathy before that's a polite thing to do :D

PR #817, opened 2020-06-23
Add the Rod of Tar Valon 

RikerW:
Opposite of Callandor for gender, serves the same purpose. 

* lawful ivory (bone) wand of striking
* Grants double spell damage & eregen when wielded
* Uses wand of orcus zap rules, since they're built-in for all artifact wands

Doesn't do half spell damage like callandor or have good base damage since it doesn't lower sanity.

----

To be honest - I have no clue how the balance on this is. Same with Callandor. Recently i've just ended up using C for artifact donations, and this is a wand of striking for consistency with spawn location. I'm not convinced letting them spawn there is a good thing, but Callandor at least should stay as a crystal sword? (rod can become a wand of light). It is quite thematic that it spawns in a tower though.

Why did I make this? I thought it was dumb that Callandor is still in there being sexist and there's no equality. Not sure adding a new arti is the best solution but w/e.


NeroOneTrueKing:
Tangentially related, we could disable Callandor spawning in the tower of law, actually. `select_floor_artifact()` could easily have a special case for that location.

I'm a little concerned that the Rod of Tar Valon would be very common everywhere, considering how relatively common wands of striking are (like Soulmirror and plate mail). Might be fine, too 🤷

RikerW:
You know, I hadn't actually thought about uh the fact that wands of striking are common. That's probably ok IFF monster starting random attack items can't be artifacts. I can't tell if 2x spell damage + eregen is better than good AC/DR + reflect + drain res or not, for purposes of "is it unbalanced that this spawns a lot".

NeroOneTrueKing:
Items in monster inventories generally are not eligible to turn into artifacts, but Striking is the 2nd most common type of wand (after Light).

2x Spell Damage + ERegen is huge for a caster and this would probably stay on their character until ascension/death. Soulmirror is good but players frequently replace it with DSM, or have already wished for DSM when they find it and go "oh well whatever". 

Personally, I'd rather see the Rod of Tar Valon (and Callandor as well) be rare and powerful than common and meh.

Chris-plus-alphanumericgibberish:
Adding an artifact breaks save compatibility, so this needs to be submitted to a dev branch not a compat branch.

Chris-plus-alphanumericgibberish (changes requested):
Would break save compatability

Issue #828, opened 2020-06-24
Astaroth's Assembly nerfs existing protection


mackeyth:
Using Astaroth's Assembly sets your protection spell timeout delay to 5 turns even if does not provide any protection.

PR #850, opened 2020-06-26
Astaroth's Assembly protection timeout

mackeyth:
Changes:
Astaroth's Assembly only grants protection if it would give more than the player already has.
AA does not change protection timeout unless it grants protection.
New message for cases where the player would have received protection from AA, but already had equal or better protection from another sourse (most likely spell of protection).

### DIFF
--- a/src/spell.c
+++ b/src/spell.c
@@ -2115,10 +2115,21 @@ spiriteffects(power, atme)
 						break;//break loop
 					}
 				}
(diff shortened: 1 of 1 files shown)


mackeyth:
Should be linked to issue #828. Either I don't know how to do that, or I don't have permission to do that.

NeroOneTrueKing (approved):
Looks good, thanks!

Fixes #828.

PR #857, opened 2020-06-28
Rebalance spellbooks

### LINKED ISSUE #679: Steal xnh's spellbook re-reading
> Spellbooks can be reread at any time. If you have more than 2000 turns left of remembering the spell, it will prompt you to make sure you really want to.

### LINKED ISSUE #859: You shouldn't be given the option to study the wardings on the Book of the Dead
It seems very weird.

RikerW:
Notable changes:

* delay is now 10*lvl, and the delay slot of the objects.c macro has been removed and mapped to level. delay is still checked in the spell.c code, since it doesn't matter which and in case of a future instant-spell-read intrinisic, or something like that. idk just didn't see a point bothering to remove it, not like it's vanishing from obj->oc_delay anyway.
* Failure rates modified. Base is still 4+int+ulvl/2 - bklvl*2. Modifiers are blessed (+10, down from 100% success), cursed (-10, up from 0% success), ward (+10), lenses (+2). New modifier: current spell memory. If the spell is known by you, you get a modifier that scales linearly up to +10 with the spell memory (so 90% memory is +9).
* Failure effects modified. The format is still rn2(lev), where lev -=1 for being ward. Also decrements effective lev now if the spell is known in general, so a re-read that fails will have nicer effects. The effects are changed to (in order of diff, 0-6 inclusive):
  * 0: teleport you with 3/4 chance, unbless the book / curse the book, if cursed always teles.
  * 1: agg monster.
  * 2: blind for rn1(100, 250) turns (250+rnd(100).
  * 3: book bites you, dealing rn1(5, 3) damage.
  * 4: freezes you for rn1(lvl*2, lvl*2)
  * 5: contact poison, unchanged.
  * 6: explosive energy, messaging changed a tad and now checks 1/2 spell damage because it's funny
* Spellbooks can now be read when you still have memory. Will warn at 10%+ and prompt yn before restoring fully. Infinite spells will always refill fully with no warn prompt.
* Minor change: spellbooks now blank at the end of the reading that sets their charges to 0, instead of the reading after that (which will fail).


NeroOneTrueKing:
What happens if I write:
Fixes #679 
Fixes #859 
?

Doesn't seem to work :(

src/spell.c
```diff
@@ -160,72 +158,105 @@ char ilet;
...
 	int lev = objects[bp->otyp].oc_level;
+
+	boolean was_in_use;
+	boolean already_cursed = bp->cursed;
+	int dmg = 0;
+	
+	if(RoSbook == STUDY_WARD) lev -= 1; //The wardings partially protect you from the magic contained within the book.
 	
-	if(RoSbook == STUDY_WARD) lev -= 1; //The wardings paritally protect you from the magic contained within the book.
+	for (int spell = 0; spell < MAXSPELL; spell++)
```
NeroOneTrueKing, line 169: It lets you declare `int spell` inside the `for()` loop? I thought the compiler didn't like that.

src/spell.c
```diff
@@ -160,72 +158,105 @@ char ilet;
 
 /* TRUE: book should be destroyed by caller */
 STATIC_OVL boolean
-cursed_book(bp)
-	struct obj *bp;
-{
+cursed_book(struct obj *bp){
```
NeroOneTrueKing: The standard for parameters is to put only their name here, and put their types below:
```
boolean
cursed_book(bp)
struct obj * bp;
{
```

src/spell.c
```diff
@@ -160,72 +158,105 @@ char ilet;
...
-			if (uarmg->greased) {
-			    grease_protect(uarmg, "gloves", &youmonst);
+		case 0:
+			if (already_cursed || rn2(4)) {
+				You_feel("a wrenching sensation.");
+				tele(); /* teleport him */
+			}
+			else {
+				if (bp->blessed) {
+					pline_The("book glows brown.");
```
NeroOneTrueKing: I believe it's technically possible to read a spellbook while hallucinating, so this could instead call hcolor:
`pline_The("book glows %s.", hcolor("brown"));`
And similarly for glowing black.

src/spell.c
```diff
@@ -160,72 +158,105 @@ char ilet;
...
-		if(Antimagic) {
-		    shieldeff(u.ux, u.uy);
-		    pline_The("book %s, but you are unharmed!", explodes);
-		} else {
-		    pline("As you read the book, it %s in your %s!",
-			  explodes, body_part(FACE));
-		    losehp(2*rnd(10)+5, "exploding rune", KILLED_BY_AN);
-		}
+			
+			if (Half_spell_damage) dmg = (dmg+1/2);
```
NeroOneTrueKing, line 247: For extra fun, why not also check `u.uvaul_duration` and reduce damage taken again? 😄

NeroOneTrueKing (changes requested):
I've just got some nitpicks; this looks good!
You don't have to make changes, they're just things to consider.

src/spell.c
RikerW: K&R style func decs are not used by anybody except for consistency with existing code. I'm not particularly concerned about that, and I don't really feel like keeping it k&r forever is a good idea anyway ([see here, which is my token anti-K&R link](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=48091)).

RikerW (commented):
Changes made to address the issues brought up.

NeroOneTrueKing (approved):

PR #874, opened 2020-07-01
Adjust fluorite octet

### LINKED ISSUE #855: Fluorite Octet naming is way off
`a blue gems named The  Fluororite Octahedron!`

### LINKED ISSUE #870: Fluorite Octet + Android interaction
Throwing the Flurorite Octet by hand as an Android throws all 8, which then return to the hand. Even by Android and Anachrononaut DPS standards, this is too much.

RikerW:
When un-id'd: "glowing cobalt eight-sided die" (that pluralizes correctly to "dice", too!)
When id'd: "fluorite octet/octahedra/octahedron"
When in a lightsaber: "burning cobalt" (instead of "burning cerulean")

Also, only throws 1 per throw when used by an android with returning.

### DIFF
--- a/include/artilist.h
+++ b/include/artilist.h
@@ -914,7 +914,7 @@ A("Plague",				BOW,							(const char *)0,
 
 /*Needs encyc entry*/
 /* die size set to 1d8 in weapon.c */
-A("The Fluorite Octahedron",		BLUE_FLUORITE,		(const char *)0,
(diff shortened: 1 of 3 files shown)


NeroOneTrueKing (approved):
Pulled and tested, and added the plural case to vtense to handle "7 Fluorite Octohedra **hit** the ogre!".
Looks great!

PR #890, opened 2020-07-02
Add 3.6 #tip command

RikerW:
Tips items out of a container. Can be used to tip items on the floor. When used on a bag of tricks or horn of plenty, empties entire contents at once. Does not require a free hand.

Bolted magic chests can't be tipped. Non-bolted can be tipped.


NeroOneTrueKing (approved):

NeroOneTrueKing:
Wait hold on. What happens when you `#tip` an unbolted magic chest, or a locked box?

RikerW:
```You see here a chest, tip it? [ynq] (q) It's locked.```  (magic chests say the same, and once they've been unlocked they're bolted ideally)

PR #900, opened 2020-07-03
Macro Magic: applied on artilist

NeroOneTrueKing:
Usage of the macro magic:
Say you want to write:
`int mynums1[] = {0, 0, 0, 0, 0, 0};`
There are 6 things to set. You'd then use SET06:
`int mynums1[] = {SET06()};`
And SET needs to have defined what default values to fill into each:
```
#define DEF01 0
#define DEF02 0
#define DEF03 0
#define DEF04 0
#define DEF05 0
#define DEF06 0
```
If you want to initialize the 4th element of the array as `CLR_BLUE`,
`int mynums2[] = {0, 0, 0, CLR_BLUE, 0, 0};`
We pass the index of the array we want to set to SET: (extra spaces added for clarity)
`int mynums2[] = {SET06(     C4(CLR_BLUE)     )};`
If the 4th index of the array is always colour, it probably makes sense to make a macro to hide the 4th index part:
`#define SET_COLOR(x) C4(x)`
So we can write:
`int mynums2[] = {SET06(     SET_COLOR(CLR_BLUE)     )};`


NeroOneTrueKing:
How macro magic works:


1) The Goal
I want to be able to instantiate a structure without directly specifying every bit:
```
struct mythingA = {"name", 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x8|0x1, 0};
struct mythingB = {/*something easier to read*/};
```

2) What does SETnn do?
SET01 sets exactly 1 parameter:
```
struct mythingA = {a1};
struct mythingB = {SET01(/*something*/)};
```
SET03 sets exactly 3 parameters:
```
struct mythingA = {a1, a2, a3};
struct mythingB = {SET03(/*something*/)};
```

3) How does SETnn expand?
```
SET03(args)
-> SET(03, args), SET02(args)
-> SET(03, args), SET(02, args), SET01(args)
-> SET(03, args), SET(02, args), SET(01, args)
```

4) What does SET do?
SET(nn, args) fills exactly one field, the nnth.
```
pline("%d  %d  %d",
       SET(01, args),
       SET(02, args),
       SET(03, args));
```
If args contains an argument that is supposed to fill the nnth field, it uses that. Otherwise, it uses a default value.

5) How can SET set a specific field?
SET(nn, args) knows it has to grab an argument that is somehow labeled to fill the nnth field, and ignore all other args.
If each argument is also numbered, we just have to match field# to arg#:
```
f   arg #
i   X| | 
e   -+-+-
l    |X| 
d   -+-+-
#    | |X
```
Rows are the field number (Rnn)
Columns are the argument number (Cmm)

6) How does SET expand?
It expands into:
```
MAYBE(defaultnn, CALL_MACRO_X_FOR_EACH(CONCAT, Rnn, args))
```
The way MAYBE is written, it effectively happens *after* CALL_MACRO_X_FOR_EACH.
For now, just know that if we're filling a field that we have an argument for, MAYBE will do nothing at all.

We'll also skim through CALL_MACRO_X_FOR_EACH.
Similarly to how SETnn works as a chain, CALL_MACRO_X_FOR_EACH uses a chain to call CONCAT with first argument Rnn and second argument 1 of args:
```
/* args = C01(x), C02(y), C03(z) */
CALL_MACRO_X_FOR_EACH(CONCAT, R01, C01(x), C02(y), C03(z))
-> CONCAT(R01, C01(x)) CONCAT(R01, C02(y)) CONCAT(R01, C03(z))
```
CONCAT concatenates the labels Rnn with each argument:
```
CONCAT(R01, C01(x)) CONCAT(R01, C02(y)) CONCAT(R01, C03(z))
-> R01C01(x) R01C02(y) R01C03(z)
```
This is the order-of-n^2 part. Each and every RnnCmm(X) combination is actually a #define.
Only the ones where `nn == mm` turn into (X), all others turn into nothing.
```
R01C01(x) R01C02(y) R01C03(z)
-> x
```

7) What if no arguments are given that match the field number?
This is where MAYBE comes in, and replaces nothingness:
```
R01C02(y) R01C03(z)
->
```
with a default value:
```
MAYBE(defaultnn, )
-> defaultnn
```

8) What if >1 arguments are given for 1 field number?
ie, `SET04( C02(x), C02(y) );`
This will not work, much like writing `int mynum = 5 7;` will not work.

Chris-plus-alphanumericgibberish (approved):
I think you should put some form of "how this works/how to use this" documentation in a comment at the top of nerosmacromagic.h Probably also a quick note at the top of artilist.h

Currently, I don't think that this should be widely used in the sources, but it is superior in cases where we would otherwise create a large number of numbered macros, like the Prop0() etc. macros.

NeroOneTrueKing:
I think the only other places this could be used are monst.c and objects.c, and I don't think it's needed at all in monst.c.

PR #920, opened 2020-07-06
Feature: abilities menu

### LINKED ISSUE #906: Add a(b)ilities menu
Abilities should be the top menu, talking the place of monster a(b)ilities as an alias for #monster.

All the various PC abilities should be accessible from this menu (spells, powers, techniques, styles, monster powers, etc).

NeroOneTrueKing:
Default bound to `B` and `ctrl-b`.

Default shows player powers and monster powers.
Can be configured to show just player powers, with a menu option for monster powers (ie \#monster)


Chris-plus-alphanumericgibberish (approved):
I'm worried that things will start colliding with both player and monster powers in the same menu, however, it hasn't happened yet and it's an easy enough fix if it does.

PR #937, opened 2020-07-07
Overhaul god_gives_benefit, remove artifact/fancy wep check from goat sizzling

RikerW:
Add 4 new god_gives_benefit options. The list now has 6 options total, chosen by rn2(6):
- uncursed restore ability & uncursed gain ability
- (old) +1 to wielded weapon, up to a max of +7. (new) if can't enchant uwep, enchants uswapwep as well, then boots, then gloves, then left/ring right. Weapons go to +7, gloves/boots/rings go to +5 regardless of +7able status. To enchant boots or gloves, you must be a martial artist either wearing artifact gloves/boots, or be a martial artist with no wielded weapon.
- identify an item that's worn/wielded, first down the list of `uwep/uswapwep/uamul/uleft/uright/uarmc/uarm/uarmu/uarmh/uarmg/uarmf`
- gives a random resistance for rn1(1000, 500) (only one that you don't have). first of poison (which is perma instead), sleep/fire/cold/shock.
- repairs an item, choosing the most damaged. in case of ties, follows uwep/uswapwep/uarmc/uarm/uarmu/uarmh/uarmg/uarmf/uarms. repairs by one level each time, if there's nothing to fix will start to erodeproof.
- (old) bless an item, going `uwep/uswapwep/uarmg/uarms/uarmc/uarm/uarmh/uarmf/uarmu/uleft/uright/uamul`. (new) adjusted to do amulets before rings. instead of always blessing, will now try to make it a BUC you don't dislike. 
  - if you hate both holy/unholy (polyselfed auton, rilmani), makes it uncursed. 
  - if you hate holy (vampire PC), makes it cursed. it's assumed anybody who hates holy is either weldproof or doesn't care about curses for some other reason
  - if you hate unholy (polyselfed angel), makes it blessed.
  - if none of the above, makes it blessed. 

Goat acidifying: 
Removes the need for the wielded item to be an artifact/oprop weapon/future weapon. Now, the wielded item needs to be: a weapon, a weapontool, or an artifact. It is intentional that a non-weapon artifact that is wielded can become sizzled, since any non-weapon artifact you're likely to be wielding is likely to be something you're hitting with or throwing anyway. 
In addition, if you are a martial artist wearing artifact gloves or boots, it will attempt to sizzle those automatically, without having to wield them (useful for monks, mostly).


src/pray.c
```diff
@@ -3946,66 +3946,235 @@ aligntyp alignment;
...
+				else if (uarmc && not_fully_identified(uarmc)) identify(uarmc);
+				else if (uarm && not_fully_identified(uarm)) identify(uarm);
+				else if (uarmu && not_fully_identified(uarmu)) identify(uarmu);
+				else if (uarmh && not_fully_identified(uarmh)) identify(uarmh);
+				else if (uarmg && not_fully_identified(uarmg)) identify(uarmf);
+				else if (uarmf && not_fully_identified(uarmf)) identify(uarmf);
+				else if (uarms && not_fully_identified(uarms)) identify(uarms);
+				else {
+					for(otmp=invent; otmp; otmp=otmp->nobj)
+					if (not_fully_identified(otmp)) identify(otmp);
```
NeroOneTrueKing: Ahem, this identifies the rest of your inventory, not a single item.
```suggestion
					if (not_fully_identified(otmp)) {
						identify(otmp);
						break;
					}
```

src/pray.c
```diff
@@ -3946,66 +3946,235 @@ aligntyp alignment;
...
+						else
+							uarmf->oeroded = 0;
+						
+						uarmf->rknown = TRUE;
+						break;
+					} 
+				}
+				break;
+			case 5: // bless/curse an item
+#define wrongbuc(obj) ((hates_unholy(youracedata) && hates_holy(youracedata)) ? \
```
NeroOneTrueKing, line 4194: I would prefer if in-function `#define`s are `#undef`ed at the end of their usage scope, so they aren't active in other parts of the file.

src/pray.c
```diff
@@ -3946,66 +3946,235 @@ aligntyp alignment;
...
+				else if (uamul && wrongbuc(uamul)) otmp = uamul;
+				else if (uleft && wrongbuc(uleft))  otmp = uleft;
+				else if (uright && wrongbuc(uright)) otmp = uright;
+				else {
+					for(otmp=invent; otmp; otmp=otmp->nobj)
+					if (wrongbuc(otmp)) break;
+					return;
+				}
+				if (hates_unholy(youracedata) && hates_holy(youracedata))
+					otmp->cursed = otmp->blessed = 0;
```
NeroOneTrueKing: There are specific unbless and uncurse functions that should be used here. Some items need the function to be called to ensure they update properly (ex, bags of holding, luckstones).

src/pray.c
```diff
@@ -3946,66 +3946,235 @@ aligntyp alignment;
...
-			    Your("%s %s.",
-				 what ? what :
-				 (const char *)aobjnam (otmp, "softly glow"),
-				 hcolor(NH_AMBER));
-			break;
+				}
+				
+				if(!(HSleep_resistance)) {
+					You_feel("wide awake.");				
+					if((HSleep_resistance & timeout) < TIMEOUT) {
```
NeroOneTrueKing, line 4026: You already checked `!(HSleep_resistance)`. 
Instead of:
```
					if((HSleep_resistance & timeout) < TIMEOUT) {
					long timer = (HSleep_resistance & TIMEOUT) + timeout;
						HSleep_resistance &= ~TIMEOUT; //wipe old timer, leaving higher bits in place
						HSleep_resistance |= timer; //set new timer
						break;
					} else {
						HSleep_resistance |= timeout;
						break;
					}
```
Do:
```
HSleep_resistance |= timeout;
break;
```

And similar for the other resistances.

NeroOneTrueKing (changes requested):
My opinion is that some of these benefits are too strong for what they are.

`god_gives_benefit` is a totally free boon granted decently often when sacrificing corpses (as in, at the same time as you are trying to get artifact gifts). ~~Restoring all injured attribute scores especially stands out as too good, considering a full boon you have to pray for has that as its primary effect (golden glow).~~ Edit: noticed it only does one stat, oops.

src/pray.c
RikerW: Intentional, I intend to unify the intrinsic-giving code into a single function, right now all code that does that uses ugly copy-pastes from eat.c.

PR #956, opened 2020-07-12
Drop bones items on death correctly

RikerW:


src/bones.c
```diff
@@ -190,8 +192,8 @@ struct obj *cont;
 			(void) add_to_minv(mtmp, otmp);
 		else if (cont)
 			(void) add_to_container(cont, otmp);
-		else
-			place_object(otmp, u.ux, u.uy);
+		else if (x >= 0 && y >= 0)
```
Chris-plus-alphanumericgibberish, line 195: This should use isok() instead of checking if it's positive

Chris-plus-alphanumericgibberish (approved):
If you don't fix the isok() thing before I merge, I'll fix it.

src/bones.c
Chris-plus-alphanumericgibberish: Merged and resolved

PR #960, opened 2020-07-14
Add more tip interactions

RikerW:
* Can tip uarmh if it's not welded to you for a response back. Peacefuls without hats wave, peacefuls with a hat they can't tip (cursed) try and fail to tip it at you, peacefuls with a hat tip it back. Hostiles will generally curse at you, unless they're invisible, in which case they're giving their current chat msg.

* Can tip a rothe to wake it up, but knocking it over. 
* Can tip a shk to offer them gold. 
* Tip a leprechaun to bribe them to go somewhere else, pacifying them.
* Tip a migo philosopher to offer them a penny for their thoughts. 
* Tipping a mercenary will prompt to throw gold if you want to bribe.

### DIFF
--- a/src/pickup.c
+++ b/src/pickup.c
@@ -3389,20 +3389,34 @@ tipmonster:
 				}
 			}
 		} else if (mtmp->isshk){
-			if (u.ugold){
+			if (!can_respond(mtmp))
+				pline("%s doesn't react.", Monnam(mtmp));
+#ifdef GOLDOBJ
+			else if (money_cnt(invent))
+#else
+			else if (u.ugold){
+#endif
 				You("offer %s a tip for their excellent service.", Monnam(mtmp));
 				if (strcmp(shkname(mtmp), "Izchak") == 0){
 					pline("Izchak thanks you for your generous tip.");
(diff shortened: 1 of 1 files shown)


src/pickup.c
```diff
@@ -3207,258 +3208,344 @@ static const char tippables[] = { ALL_CLASSES, TOOL_CLASS, 0 };
...
+						pline("The creature has fallen and can't get up.");
+					else
+						You("successfully tip the rothe onto its side, startling it awake!");
+					
+					mtmp->msleeping = 0;
+					mtmp->mfrozen = 1;
+				}
+			}
+		} else if (mtmp->isshk){
+			if (u.ugold){
```
Chris-plus-alphanumericgibberish: This doesn't seem to have a can_respond(mtmp) check?

Also, dnethack has otherwise maintained compatibility with gold objects for some reason, which is not the case here.  I don't think it would be hard to adjust it.

Chris-plus-alphanumericgibberish (changes requested):

Chris-plus-alphanumericgibberish (approved):

PR #998, opened 2020-07-23
Specify while-blind name of objects

### LINKED ISSUE #854: Objects have colors while blind and un-identified
Examples: 
* `blue-glowing vibrosword` (this one is really funny)
* `curved black and gold sword`
* `hard black handmirror`
* `silver pellet`
* `spider-emblemed crossbow` (unless this is inset?)

etc. armor has this problem too

Solutions would either be to make blindness use an intelligent un-identified description, but honestly that's probably overkill. Just make obj->app never refer to colors or anything like that.

NeroOneTrueKing:
NOTE: based on #996 to avoid merge conflicts (first 5 commits)

Example, a "white headscarf" is shown as a "headscarf" when you don't know the objects `dknown`.
View f493537 to see all special cases made.

Does not touch lightsaber blade colors, because Chris said he was poking at them.

Blind names are not valid to use to wish for an object. This should probably change in the future.

### DIFF
--- a/include/objclass.h
+++ b/include/objclass.h
@@ -61,9 +61,13 @@ struct objclass {
 
 #define	UPPER_TORSO_DR 0x01 /* body armor, shirt, cloak (2x weight) */
 #define	LOWER_TORSO_DR 0x02 /* body armor, cloak (2x weight) */
-#define	HEAD_DR        0x04 /* helmet, cloak */
+#define	HEAD_DR        0x04 /* helmet */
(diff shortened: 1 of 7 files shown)


NeroOneTrueKing:
Also doesn't fix artifact descriptions, (esp Fluorite Octet). That will likely be done via significantly modifying artifact descriptions to categorize adjectives.

Chris-plus-alphanumericgibberish (approved):

PR #1025, opened 2020-07-31
Grandmaster's robe isn't a good first gift

NeroOneTrueKing:


Chris-plus-alphanumericgibberish:
There's nothing that makes them a guaranteed first gift, right?

I'm inclined to leave Hammerfeet in the pool but take out Grandmaster's Robe (since there are so many other ways to get it)

NeroOneTrueKing:
The `ARTG_GIFT` flag indicates that if a race/role has no guaranteed first gift, they should draw an artifact from the standard pool of eligible artifacts that also have the `ARTG_GIFT` flag.

Sure, a player could make do with Hammerfeet. The Grandmaster's Robe doesn't fit here though.

Chris-plus-alphanumericgibberish (approved):

PR #1055, opened 2020-08-04
Allow wishing for figurines of uniqs/nowish monsters

RikerW:
The attempt to create the monster will just fail, if it would have previously been nowish. Is this rude to players who wished for a figurine of <x>? Yeah, but they wouldn't be getting the figurine anyway. The intention is that if you really want a swag figurine of demogorgon to ascend with, then by jove you'll get one!

Diff is cluttered because I actually like, indented this function.

### DIFF
--- a/src/dog.c
+++ b/src/dog.c
@@ -98,39 +98,43 @@ boolean quietly;
 	struct monst *mtmp = 0;
 	int chance, trycnt = 100;
 
+
 	do {
 	    if (otmp) {	/* figurine; otherwise spell */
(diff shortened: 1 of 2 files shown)


NeroOneTrueKing:
Fair enough. Wishing for a "figurine of demogorgon" before would give you something like a figurine of a jackal, it's not like this makes you more able to accidentally waste a wish.

NeroOneTrueKing (approved):
Looks good!

PR #1059, opened 2020-08-04
Remove shield casting penalty being tied to weight

NeroOneTrueKing:
A "tiny" dragon scale shield is now no less prohibitive than a regular one.

Effectively, this change is +2 AC for characters that decide to wear a dragon-scale shield and want to cast spells, and less munchkinry.


Chris-plus-alphanumericgibberish:
I prefer the current behavior, which explicitly ties the spellcasting penalty to the weight of the object.

PR #1080, opened 2020-08-09
Add Dirge

RikerW:
~~Stolen~~ Inspired by EvilHack Dirge.

* chaotic intelligent knight-favoring malleable long sword
  * Will always evade the grasp of non-chaotics and non-knights, you must be a chaotic knight to wield. Helms of opposite alignment are accepted for wielding purposes.
* Can only be obtained by same-race sacrifice on a chaotic altar, while a permanently chaotic knight wielding a long sword. This will convert the long sword into Dirge, rustproofing it, clearing erosion, and setting enchantment to 0 if it was previously lower. This will not uncurse it.
* Has +5 to hit and +1d10 acid damage
* Has +6 unholy damage when cursed
* Permanently coated in acid poison
* Grants acid resistance when wielded

This is mostly just a feature suggestion I wanted to get out there. It seems like fun, if not particularly balanced, and shouldn't be too busted since it requires a permanent conversion to chaotic to get. Should be only marginally stronger than Excalibur in general, and much less useful in the late game considering the extra d10 acid vs 3d7 holy. Thoughts?

### DIFF
--- a/include/artifact.h
+++ b/include/artifact.h
@@ -453,7 +453,7 @@ extern struct artinstance artinstance[];
 				)\
 			)
 /* artifact has no specific material or size, eg "silver Grimtooth" */
-#define is_malleable_artifact(a) (is_nameable_artifact((a)) || (a) == &artilist[ART_EXCALIBUR])
+#define is_malleable_artifact(a) (is_nameable_artifact((a)) || (a) == &artilist[ART_EXCALIBUR] || (a) == &artilist[ART_DIRGE])
 
 #define is_living_artifact(obj) ((obj)->oartifact == ART_TENTACLE_ROD || (obj)->oartifact == ART_DRAGONHEAD_SHIELD || (obj)->oartifact == ART_CRUCIFIX_OF_THE_MAD_KING || (obj)->oartifact == ART_RITUAL_RINGED_SPEAR)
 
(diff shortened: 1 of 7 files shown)


Chris-plus-alphanumericgibberish:
Could give it the self-acidifying property instead of a permanent coat of acid poison.  Self-acidifying means that the first few hits of every fight have the extra acid damage, then it drops back to +1d10 only.

Edit: Is that acid AND poison coatings, or just an acid coating?

Edit 2: Note that Excal is... 5.5 physical + 10.5 holy for +16 basic, while this version of dirge is 5.5 acid + 5.5 acid + 6 unholy, for +17 basic, so a permanent acid coating is not unbalanced. The player also spends more of the game fighting holy-hating monsters than they do fighting unholy-hating monsters.

RikerW:
The intention was only acid poison, but I'm considering making it poisoned instead (mainly since that much acid damage is kinda weird). Poison is definitely a nerf for endgame damage, but does have the instakill vs some foes. I also considered filth, but that's stupid as a permanent due to the instakill chance. 

Yeah, the average damage is supposed to be equal to or possibly a tad lower than excalibur. It may need to go lower still, considering this can be drooled on, but I'm not sure yet.

NeroOneTrueKing (changes requested):
As written, Dirge has no bonus unholy damage, which doesn't match your design notes. Please confirm.

Edit: it looks cool and I like it

Chris-plus-alphanumericgibberish (approved):

PR #1096, opened 2020-08-12
Eating an amulet of restful sleep will cause you to fall asleep once.

### LINKED ISSUE #666: Eating amulets of restful sleep is odd
Eating an amulet of restful sleep has effects similar to wearing it -- every 1-100 turns, you fall asleep.

However, the amulet_on() and amulet_off() code overrides this; after eating one amulet, putting on and taking off another amulet of restful sleep will purge your cycle of sleepiness.

NeroOneTrueKing:
Fixes amulet on/off solution... by making it not necessary.


Chris-plus-alphanumericgibberish:
I think the problem is the odd method of curing the restful sleep intrinsic, rather than the intrinsic itself.

PR #1100, opened 2020-08-12
Add waterproofing as a proper trinsic

RikerW:
Obtainable from white dragon scales, faceplates, oilskin cloaks for items. Dividing word now is checked in there, as is preservation. It does NOT include greased cloaks, since those have handling to possibly wear off grease, so those are implemented in the appropriate places after the checks. 

Enki grants the waterproof property in sounds.c like he should, given the new system.


NeroOneTrueKing (approved):
Looks good

PR #1106, opened 2020-08-14
Fix ssuffix buffer management

NeroOneTrueKing:
Why the heck wasn't this breaking in an obvious manner?


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
Re: why not breaking in an obvious manner: both ways set aside `BUFSZ * SSUFFIX_BUFFERS chars`-worth of memory, so it shouldn't overflow or anything.  I think things would often end up partially overlapping when multiple suffixes were used in one print, but I think that's rare.

PR #1108, opened 2020-08-14
Fix findtravelpath() going back and forth

### LINKED ISSUE #1015: Travel command can get stuck in infinite loops
Not sure how to reproduce yet, but we keep getting reports of this every now and then.

NeroOneTrueKing:
This happened when the following occurred:
1) you tried to travel to somewhere you did not have a clear path to (blocked off, or unexplored tiles in the way)
2) there was an intermediary location, closer to your target, that *could be reached*, in LoS
3) the path to reach that intermediary location brought you further away from your original target AND broke LoS to your intermediary location

Simply dropping the LoS requirement fixes findtravelpath(), which is what this first commit does.

I am exploring the possibility of re-enforcing the LoS requirement when picking intermediary locations, but saving a single intermediary location to `u` so that it can only pick better intermediaries (because LoS changes as you move).

### DIFF
--- a/include/flag.h
+++ b/include/flag.h
@@ -410,7 +410,7 @@ struct instance_flags {
 	boolean  lootabc;	/* use "a/b/c" rather than "o/i/b" when looting */
 	boolean  showrace;	/* show hero glyph by race rather than by role */
 	boolean  travelcmd;	/* allow travel command */
-	boolean  travelplus;/* if travel command should attempt to path through the unknown */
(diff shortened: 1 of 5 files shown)


NeroOneTrueKing:
This is now the basis of a discount autoexplore feature. If you set travelplus high enough, it tries to map out everywhere that could potentially get you to your target location if it can't find a path.

Chris-plus-alphanumericgibberish (approved):

PR #1121, opened 2020-08-16
Tweak armor weight-types

NeroOneTrueKing:
1) Specify in objects.c (using the size field)
2) Lighten several


Chris-plus-alphanumericgibberish (approved):
Armor encumbrance needs to be looked at.

PR #1125, opened 2020-08-17
Replace Gae/Tulani in mklev spawns appropriately

RikerW:
This fixes a game with Gae Eladrin instead of Tulani still spawning Tulani on the plane of water, and any other hypothetical desfile inclusions. It assumes G_NOGEN is used to track which should spawn in that game.

### DIFF
--- a/src/makemon.c
+++ b/src/makemon.c
@@ -7571,6 +7571,13 @@ register int	mmflags;
 	)
 		mmflags |= MM_EDOG;
 
+	/* replace desfile Tulani with Gae, and vice versa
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing (commented):
Looks good to me, but I'm never certain what Chris wants alt Eladrin, so I'll leave it to him to fully approve.

Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
The alt castes should be replaced in most situations, except when loading bones. The random roll is determining which caste is rare and which is more common (high-caste outsiders never being very common in the first place).

PR #1131, opened 2020-08-17
Confuse monster can only be applied in melee

NeroOneTrueKing:


Chris-plus-alphanumericgibberish:
This was a buff to confuse monster.

PR #1137, opened 2020-08-18
add compound option delay_length:normal|short|instant

bubblehack:
I would have done an absolute time value but some of the window systems have different "normal" delays so changing that seems like a bad move for people who are used to it.


bubblehack:
[preview](https://i.imgur.com/WsVxAcm.gifv)

Chris-plus-alphanumericgibberish (approved):

PR #1142, opened 2020-08-22
Rewrite how macromagic works

NeroOneTrueKing:
Now operates in O(args) time, instead of O(args^2).

Compile time for objects.c is down to about 5 seconds on my computer, down from what felt like forever (~30 seconds).

Conflicts with #1091; to resolve, merge #1091 first and then manually edit its MONS() macro and add the 7x `0,`.

p.s. sorry Chris, still figuring out my new git workflow


NeroOneTrueKing:
Closing.
Got my git workflow figured out, making the PR from my fork.

PR #1150, opened 2020-08-25
Fix buffer overflow when adding cmds

amateurhourrrr:
There were missing entries for some of the extended commands that get
added if the player is in wizard mode, causing adding the debug commands
to overflow the buffer.

Additionally moves a comment such that the null entries' comments are in
consistent order with the command entries themselves.


amateurhourrrr:
noticed this is fixed in devel, my bad

PR #1151, opened 2020-08-25
Double size of nulls buffer

amateurhourrrr:
The buffer is presently too small, so the game actually has a buffer
overflow upon starting the game.

The size of the fruit struct is larger than the previous length of
40 null values, causing an overflow in save.c:1003

### DIFF
--- a/src/save.c
+++ b/src/save.c
@@ -36,11 +36,6 @@ STATIC_DCL void FDECL(savelev0, (int,int,int));
 STATIC_DCL boolean NDECL(swapout_oldest);
 STATIC_DCL void FDECL(copyfile, (char *,char *));
 #endif /* MFLOPPY */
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing:
I also think that in the case of save.c using nulls, there should be a dedicated "write X bytes with null" function that does not rely on `nulls` being at least as large as the amount being written.

amateurhourrrr:
I think you're right there, it's a pattern that crops up a lot

amateurhourrrr:
@NeroOneTrueKing I've changed strategy here to match how vanilla handles this precise case

Let me know what you think :+1: 

I've also retargetted it at devel

NeroOneTrueKing:
Looks great!

Chris-plus-alphanumericgibberish (approved):

PR #1156, opened 2020-08-25
Fix picking special rooms

NeroOneTrueKing:
1) rnd((x=0)) = 1; don't call if x==0.
2) only loop while i-->0.


amateurhourrrr:
This doesn't fix it unfortunately

I'm seeing it put really weird values in special_rooms, will take another look tomorrow

Chris-plus-alphanumericgibberish (approved):
This looks like an improvement anyway, so I'll merge it.

PR #1179, opened 2020-08-31
Deprecate "wepgone". Send (struct obj **)(&weapon).

### LINKED ISSUE #1172: use after free when weapon is destroyed during an attack
When a wielded object is destroyed during an attack, the xhity functions continue on as though the obj hasn't been freed.

This is easiest to see with a wielded potion smashed on a random monster
compile with address sanitizer to almost instantly see it

Could potentially lead to write after free in some contrived circumstances (speculating: hitting disenchanter with potion)

first line it throws when compiled with address sanitizer for the particular case with the potion is about src/xhity.c:755

NeroOneTrueKing:
Instead of adding boolean * wepgone to more things, change them to send an object pointer pointer so that the called function is able to nullify the caller's copy of the object pointer.

Minimalistic example:
```
hmon(&weapon);
if(!weapon) {
    // weapon was destroyed in hmon()
}
```


Chris-plus-alphanumericgibberish (dismissed):

Chris-plus-alphanumericgibberish:
This is a really good point: More functions that can destroy what they're given should send in pointer-pointers instead of pointers.

Chris-plus-alphanumericgibberish (approved):

PR #1187, opened 2020-09-05
Stationary engulfers stay in place

### LINKED ISSUE #1185: Stationary engulfers move when expelling the player
I made them not to do this for mvm combat, but not for mvu.

NeroOneTrueKing:
And tries to put player back where they were grabbed from.

### DIFF
--- a/src/xhity.c
+++ b/src/xhity.c
@@ -8202,6 +8202,8 @@ int vis;
 			}
 			else {
 				/* or maybe pluck the player */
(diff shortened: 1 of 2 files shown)


NeroOneTrueKing:
Edit: trying to put player back where they were grabbed from is buggy; and only has an effect if the player immediately exits the engulfer 😞

Chris-plus-alphanumericgibberish (approved):

PR #1190, opened 2020-09-06
Add new artifact: Bloodletter

RikerW:
* unaligned (iron) morning star
* +1d8 to hit and +1d12 blood damage
  * Deals extra bonus blood damage at lower health. At under 50% of maximum health, deals an additional +1d12 for a total of +2d12. At under 25% of maximum health, deals an additional +1d12 for a total of +3d12.
* When invoked, self-impales (dealing damage equal to 20% of your max hp, can be lethal) to coat in blood. This lasts for 20 + 2z9 turns. While coated in blood:
  * Appears as "sanguine" in your inventory. 
  * Damage dealt is also treated as energy damage, in addition to being blunt and piercing.
  * Deals an extra flat +defender level damage to monsters with blood. 
  * Hitting high level monsters during this time can decrease the amount of time remaining on the blood coating. It subtracts `defender_level/10 - rn2(5)` (rounded down) from the time remaining, and never adds more time.
  * Can be applied to slam down, releasing the tainted blood. This ends the active effect but causes a 6d6 rotten blood explosion. The explosion only hurts creatures with blood, and also deals a flat +defender level damage to affected creatures.

### DIFF
--- a/include/artifact.h
+++ b/include/artifact.h
@@ -293,6 +293,7 @@ struct artinstance{
 #define RoSPkills avar1
 #define BoISspell avar1
 #define RRSember avar1
+#define BLactive avar1
 	long avar2;
 #define SnSd2 avar2
 #define RoSPflights avar2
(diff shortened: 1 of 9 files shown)


NeroOneTrueKing:
~~Corrections:~~
* When under 50% hp, deals an additional 1d12 damage to creatures with blood (stacking with previous 1d12)
* When under 25% hp, deals an additional 1d12 damage to creatures with blood (stacking with previous 2d12)
* * Bloodletter does not multiply its base damage or enchantment.
* Invoke duration is `20 + [0-9] + [0-9]` turns; not `20+2d10`.

Edit: main description has been updated with these corrections.

Chris-plus-alphanumericgibberish (approved):

PR #1198, opened 2020-09-13
Sickles are throwing weapons

NeroOneTrueKing:
And move that #define to obj.h where it belongs.

### DIFF
--- a/include/obj.h
+++ b/include/obj.h
@@ -595,6 +595,27 @@ struct obj {
 			 objects[(otmp)->otyp].oc_skill == -P_FIREARM)
 //#endif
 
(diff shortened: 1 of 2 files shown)


Chris-plus-alphanumericgibberish:
Sickles in general aren't intended to be throwing weapons, and a lot of stuff spawns with sickles that they're supposed to use in melee rather than chuck at you. Is this going to interfere with that?

NeroOneTrueKing:
Oh it's just Sickle Moon that's intended to be a throwing weapon? Oops.

This change does not add sickles to the "monsters will throw these" list, which is in weapon.c.

Chris-plus-alphanumericgibberish (approved):

PR #1207, opened 2020-09-23
Enhanced disintegration can be reflected normally

RikerW:
So amulets of reflection can block black dragon breath, among others.


NeroOneTrueKing:
I think the correct method would be to change the places where disintegration beams are being created.
ie, projectile.c line 2718, add a condition that disintegration breaths are not to be marked as "refl_advanced".

NeroOneTrueKing (approved):

PR #1254, opened 2020-11-26
Fact-check the Oracle's consultations.

NeroOneTrueKing:


Chris-plus-alphanumericgibberish (approved):
"Elbereth" still works, especially for Elves.  Some checking to give the best warding advice (use Elbereth > use Lolth > use wards) would probably be good.

PR #1265, opened 2020-12-04
Add more light circles

amateurhourrrr:
With angels having a radius 7 lightsource, and certain abilities using
3x a given radius, we needed more data to avoid a buffer overflow.

This was totally eyeballed by drawing them out in a similar style to the
lesser circles and isn't particularly great

if someone wants to redo this or make the light circles actual circles
that'd be cool, but I couldn't figure out the originals were done


Chris-plus-alphanumericgibberish (approved):
I get the sense that the originals were done by eyeballing it, too.

Though maybe it's an old space-for-time trade?

PR #1268, opened 2020-12-05
Guard against segfault in check_reseed

amateurhourrrr:
Opted to go for an impossible here as it doesn't necessarily prevent
gameplay continuing without reseeding, though perhaps a panic would be
appropriate instead


Chris-plus-alphanumericgibberish (approved):
In general, I believe that NetHack is too quick to panic(), which tends to turn inocent bugs into game-destroying bugs. 

dNethack should not panic() unless there's no way for it to proceed.

PR #1272, opened 2020-12-11
Rewrite obj_to_glyph as function

amateurhourrrr:
This rewrites obj_to_glyph to avoid the unusual use of a global variable
to store an integer

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -249,6 +249,22 @@ map_trap(trap, show)
     if (show) show_glyph(x, y, glyph);
 }
 
+int
+obj_to_glyph(obj)
+    struct obj *obj;
+{
+    int otyp = Hallucination ? random_object() : obj->otyp;
+
+    if (otyp == CORPSE) {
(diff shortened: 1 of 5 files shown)


src/display.c
```diff
@@ -249,6 +249,22 @@ map_trap(trap, show)
...
 
+int
+obj_to_glyph(obj)
+    struct obj *obj;
+{
+    int otyp = Hallucination ? random_object() : obj->otyp;
+
+    if (otyp == CORPSE) {
+	int corpsenm = Hallucination ? random_monster() : obj->corpsenm;
+	return corpsenm + GLYPH_MON_OFF;
```
NeroOneTrueKing: Shouldn't the constant be `GLYPH_BODY_OFF`?

NeroOneTrueKing (changes requested):

src/display.c
amateurhourrrr: You're right, this should be corrected in the new commit :+1:

NeroOneTrueKing (approved):

PR #1273, opened 2020-12-12
Guard against null pointer in melee_polearms macro

amateurhourrrr:
This is used in hmoncore and was seen in a segfault when throwing a
polearm upwards that hit the ceiling during fuzzing


NeroOneTrueKing (changes requested):
Considering that the rest of mondata.h's macros don't guard `ptr`, I think it would be odd to do so here, and instead the caller should be careful that `ptr` exists before calling the macro.

Instead, in xhity.c (ln 11410), guard against using `pa` without first checking that it exists.

amateurhourrrr:
Closed as this has been done on devel already

PR #1274, opened 2020-12-12
Clamp circle_ptr radius to [0, MAX_RADIUS]

amateurhourrrr:
We already expanded max radius to 21, and there are two monsters for
which this ought to perhaps be expanded further up to 27

Regardless of if we do that or not, it is sensible to clamp the radius
anyway to prevent out of bounds reads.

### DIFF
--- a/include/vision.h
+++ b/include/vision.h
@@ -75,7 +75,9 @@ extern char *viz_rmax;			/* max could see indices */
  */
 #define MAX_RADIUS 21	/* this is in points from the source */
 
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing (changes requested):
Are you sure this will work correctly? Clamping `circle_ptr` will prevent reading from where `circle_data` is missing, but it won't stop the caller of `circle_ptr` from acting as though it was getting data for those large radii.

amateurhourrrr:
Ahh you're totally right, I didn't think about it fully

closing, will just have to expand the data again

PR #1277, opened 2020-12-13
Add impossible guard for Armor_off with null uarm

amateurhourrrr:
Occurred during fuzzing but I haven't a clue how to reproduce this.

Somehow, uarm was set to 0 while the armour was being taken off, which
causes a segfault on reading from low addresses

This guard avoids the segfault and reports an impossible so if a player
manages to trigger this we can hopefully get more information

### DIFF
--- a/src/do_wear.c
+++ b/src/do_wear.c
@@ -790,7 +790,11 @@ Armor_off()
 	boolean checkweight = FALSE;
     takeoff_mask &= ~W_ARM;
 	if(arti_lighten(uarmu, FALSE)) checkweight = TRUE;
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing:
Might be better to add a `static struct obj * taking_off_obj` to all of do_wear.c, set it when beginning to don/doff an object, and cry a warning when that pointer doesn't match the expected object being donned/doffed. This method would give a pointer to the missing object, which would likely help us find what happened.

amateurhourrrr:
I don't think it can be static as Armor_off occurs in steal.c as well, but that's a good idea
I'll get that done soonish so you can have a look

Chris-plus-alphanumericgibberish (approved):

PR #1333, opened 2020-12-25
Re-set potion colour after changing potion->otyp

NeroOneTrueKing:
Note that constantly resetting obj->color is *not* okay in some cases right now -- that should probably get changed so that they are MOD obj->oid and not rn2()


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
It should not be expected that constantly resetting obj->color is ok (it's not intended to be like weight, which *is* constantly reset).

PR #1347, opened 2020-12-27
Creatures with explosion attacks call their on-death explosions too

NeroOneTrueKing:
This is a bugfix to fern spores, primarily, but it also affects other spheres.
Other spheres now do their on-death explosion (AT_BOOM) *in addition* to their hitting-target explosion (AT_EXPL). Which makes sense, but is also certainly a buff to them!

Finally, fix messaging with regards to AT_EXPL monsters attacking your displaced image.


Chris-plus-alphanumericgibberish:
The idea with spheres is they do either a 4d6 focused blast if allowed to hit or a 2d6 unfocused explosion if detonated early.  Fern spores either settle peacefully or explode violently if tampered with.

PR #1355, opened 2020-12-29
Massively nerf flaying weapons

NeroOneTrueKing:
Armor destruction:
1) "lesser" flaying weapons only apply at all 1/3rd of the time
2) "greater" flaying weapons disenchant by 1, not 2-4.
3) "Thorns" disenchants by 1-3, not 2-7.
also check visibility of messages.

Movepoint loss:
1) "lesser" amount doubled for effectively 2/3rds as much when combined with above
2) "greater" amount reduced to 2/3rds as much
3) "Thorns" amount doubled to be equal to "greater".

Hopefully these changes will make being in melee with an enemy with a flaying weapon less terrible for your gear.

This comes after seeing Malcanthet destroy a fully-enchanted piece of endgame gear with each swing.

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -4547,23 +4547,28 @@ boolean * messaged;
 
 	/* flaying weapons */
 	if (check_oprop(otmp, OPROP_FLAYW)
-		|| check_oprop(otmp, OPROP_LESSER_FLAYW)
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish:
I adopted a less-massive nerf to flaying.

PR #1365, opened 2020-12-31
Double maxweight per size above medium

RikerW:
Since weapon weights are doubled, it makes sense you should be able to use appropriately sized items.

### DIFF
--- a/src/u_init.c
+++ b/src/u_init.c
@@ -2309,8 +2309,10 @@ u_init()
 			else if(isSignetRing(pobj->otyp)){
 				pobj->ohaluengr = TRUE;
 				pobj->oward = u.uhouse;
-				pobj->opoisoned = OPOISON_SLEEP;
-				pobj->opoisonchrgs = 30;
+				if(!Role_if(PM_EXILE)){
+					pobj->opoisoned = OPOISON_SLEEP;
+					pobj->opoisonchrgs = 30;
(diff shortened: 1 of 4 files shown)


NeroOneTrueKing:
I see this doesn't penalize gnomes 😄

NeroOneTrueKing (changes requested):
There are two places in weapon.c that calculate maxweight of offhand weapons -- one for damage, and one for accuracy. The offhand-weight calculation could be moved into its own function, or you can just do the modification in both places.

Gnomes notwithstanding (it's fine by me if they don't get additional penalties, I suppose).

RojjaCebolla:
just FYI gnomes are still bad

NeroOneTrueKing (changes requested):
Just noticed another issue, actually.
This commit changes offhand limits like so:
Medium = 1x
Large = 2x
Huge = 4x
Gigantic = 10x

But the weight of those items is 
Medium = 1x
Large = 2x
Huge = 3x
Gigantic = 6x

The formula:
`maxweight *= 2*(youracedata->msize - MZ_MEDIUM);`
should instead be
`maxweight *= 1+(youracedata->msize - MZ_MEDIUM);`

I'm going to ignore the unrelated commits that snuck in; hopefully they won't cause any merge conflicts.

PR #1373, opened 2020-12-31
Remove randomization of object colors

NeroOneTrueKing:
This makes set_object_color() safe to call on an object at will, like weight().
"Randomness" of colors is done by `o_id%x` instead of `rn2(x)`, so any one item will be self-consistent.


Chris-plus-alphanumericgibberish:
Color is supposed to be arbitrarily variable, the set function shouldn't be recalled constantly.

PR #1376, opened 2021-01-02
Adjust Mitre/EotOW to be appropriately sized

RikerW:
As-is, the EotOW are completely unwearable on a chiropteran monk - since you can't resize lenses/sunglasses, and yet they only fit appropriately sized characters. The Mitre is better, but it still needs a resizing & a body plan (longhead & animaloid) adjustment.  The Noble crowning helms should probably be appropriately sized, even if they would fit anyway, with the body plan adjusted as well.

I only adjusted those since they are unusable in their current state (EotOW) without adjustments (Mitre, crowning helms). Other potential artifacts would be Itlachiayaque, the Veil of Latona, Hermes' Sandals or the noble quest artifacts. I don't really think they need to be poked - no clue about Latona but I think Hermes' Sandals would be sized for Hermes' (who apparently has a medium humanoid avatar), and Itlachiayaque is an ancient artifact and usable as-is anyway. Lastly, the noble cloaks aren't even obtainable as a funkily sized race right now - chiro/gnomes can't be nobles. Though I think chiro rangers, gnomish arcs, and gnomish healers might have poorly fitting artifacts right now.

Note that you can still get crowned with the noble pantheon as a Chiropteran to receive the helms, so they're added in here as well despite not being normally obtainable. Obviously this doesn't fly for the cloaks. Amusingly enough, before this, a Chiropteran priest crowned as a noble could have the helm put on directly, bypassing the standard body plan checks, but would be unable to put it back on without adjustment if doffed.

Also added lenses/sunglasses to the "sized equipment" bit of the mkobj.c code. That doesn't include artifacts, so it's mostly irrelevant, but could definitely be noticed.

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -186,14 +186,6 @@ hack_artifacts()
  	    artilist[ART_SINGING_SWORD].role = NON_PM;
 	}
 	
-	if(Role_if(PM_MONK)){
- 	    artilist[ART_EYES_OF_THE_OVERWORLD].size = youracedata->msize;
-	}
-	
-	if(Role_if(PM_PRIEST)){
- 	    artilist[ART_MITRE_OF_HOLINESS].size = youracedata->msize;
-	}
-	
 	if(Role_if(PM_ANACHRONONAUT)){
 		artilist[ART_CRESCENT_BLADE].race = NON_PM;
 		artilist[ART_CRESCENT_BLADE].alignment = A_NONE;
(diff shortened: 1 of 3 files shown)


NeroOneTrueKing:
I just want to double-check that the changes you expect to be in these commits are all there -- there's some back-and-forth.

You mentioned adjusting the noble cloaks, which you wrote in in the 2nd commit, but your 3rd commit deleted those changes.

I'll defer to Chris on what is and isn't intended on the subject of artifact-sizes not being appropriate for the player-character.

Chris-plus-alphanumericgibberish (approved):

PR #1386, opened 2021-01-05
Nerf Octetsabers

NeroOneTrueKing:
Consider the other lightsaber artifacts: they don't really make the lightsaber much stronger, even the Annulus.
An octetsaber, at max luck, was over 2x damage compared to an unsocketed lightsaber (at max enchantment. more like 4x damage when comparing at +0).

Post-nerf, an octetsaber at max luck will be a bit more damage than an unsocketed lightsaber (average of +5 flat per hit; doubled for doublesabers).

Removes unlit octetsabers entirely. An Octahedron shouldn't boost damage if it isn't actually being used as the weapon -- putting an Octahedron into a sack won't make you deal 1d8 expl-lucky when swinging the sack around, so it shouldn't affect a lightsaber that isn't actively using its focusing gem!

### DIFF
--- a/src/weapon.c
+++ b/src/weapon.c
@@ -626,9 +626,9 @@ int otyp;
 
 	case SEISMIC_HAMMER:		if (chrgd){ ocd *= 3; } break;
 	case ACID_VENOM:			if (obj&&obj->ovar1){ ocn = 0; flat = obj->ovar1; } else{ add(6); } break;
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (approved):
I'm a little concerned about unintended interactions from replacing **spe_mult = 3** with **spe_mult \*= 3** in the lightsaber code.

If we wanted to avoid setting spe_mult, I think spe_mult += 2 would be better there than *3.

PR #1393, opened 2021-01-08
Fern spores have no blood

RikerW:


Chris-plus-alphanumericgibberish:
I have a version using is_fern_spore() on local.  Elemental spheres also need to be handled still, though.

PR #1399, opened 2021-01-10
Handle newcham() making horrors

NeroOneTrueKing:
Oddly, chameleons are allowed to be generated in the form of something nopoly, but not change their form into one. Seems like a bug.


amateurhourrrr:
it's probs nothing to be concerned about but it does duplicate some things in set_mon_data that gets called immediately afterwards

NeroOneTrueKing:
Nothing duplicated between this and `set_mon_data()`, though this is duplicate code to what is done in makemon.c when generating horrors.
What's in this patch sets up the EHOR struct for use in `set_mon_data()`, because the latter chooses to use what's in the struct instead of looking to the mons[] array (so it can handle horrors in bones files and Nameless horrors).

amateurhourrrr:
I misread it, sorry!

Chris-plus-alphanumericgibberish (approved):

PR #1402, opened 2021-01-10
Allow reading of anti-clockwise metamorphosis glyphs

RikerW:
The read.c check was not treating them as valid, due to an off-by-one error on the starting index of the glyphs.


NeroOneTrueKing (approved):
This happened when re-reversing the clockwise and anti-clockwise glyphs -- this wasn't updated. Thanks, riker!

PR #1440, opened 2021-01-23
Oops, don't put save() functions inside `if(bwrite(mode)){...}`

NeroOneTrueKing:
Those save functions handle mode-variant behaviour on their own.


amateurhourrrr:
With this it panics upon using wizmakemap, not sure what the solution is
Edit: this is no longer true

Chris-plus-alphanumericgibberish (approved):

PR #1470, opened 2021-02-02
Fix handling of Cthulhu to not affect all other widegazes

NeroOneTrueKing:
I think this is what was intended for Cthulhu's gaze attack.


Chris-plus-alphanumericgibberish:
I flipped the default when adding cthulhu's handling (cthulhu's handling is also backwards here)

PR #1494, opened 2021-04-29
Add  "Slimify" monster spell

NeroOneTrueKing:
Turns target into slime. Delayed vs player, instant vs monsters (because they don't have any handling for these delayed things).
Not on any monsters' prefered spells lists, but can rarely be chosen by the Wizard of Yendor.


NeroOneTrueKing:
On further inspection, AT_NONE AD_PHYS attacks are not yet automatically aborted. <- TODO

Chris-plus-alphanumericgibberish (approved):

PR #1519, opened 2021-05-07
Fix impossibles from wishing for partly eaten/drained foodstuffs

RikerW:
The code to set nutrition of partly eaten/drained items was improperly setting the nutrition to 1 at times when it it shouldn't be - like for tins and wraith corpses, which both have base nutritions of 0.
This would give an impossible, which is now fixed.

Also includes a minor refactoring, adding the vanilla obj_nutrition function to replace most ocurrences of `objects[otmp->otyp].oc_nutrition` / `mons[otmp->corpsenm].c_nutrit` or whatever. Since I copy-pasted it with modifactions, there's now some minor nutrition related changes (elves get +25% nutrition from lembas, orcs get -25%, and dwarves get +15% from cram).

Also fixes the long-standing bug where drained corpse wishes always ended up with 1 nutrition "over-drained" corpses (oeaten wasn't being set, leading to trying to reduce from 0 rather than the base nutrition, hitting the minimum nutrition of 1).

Note that now, a partly eaten wraith corpse wish will just fail to make it partly eaten - since there's no real way to denote a partly eaten 0 nutrition corpse in the current code. A partially drained one will just give you a 'drained' corpse that can't be drained again, but is the same nutrition, due to the isdrained field.

I would not be opposed to somebody testing the hell out of this - I spent 5 minutes or so wishing for random things and consuming them, but I'm not sure it's bug free at all. Hopefully.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -784,6 +784,7 @@ E boolean FDECL(bite_monster, (struct monst *));
 E void FDECL(give_intrinsic, (int, long));
 E void NDECL(fix_petrification);
 E void FDECL(consume_oeaten, (struct obj *,int));
+E int FDECL(obj_nutrition, (struct obj *));
 E boolean FDECL(maybe_finished_meal, (BOOLEAN_P));
 E int FDECL(ask_turns, (struct monst *, int, int));
 E int FDECL(ask_cp, (struct monst *, int));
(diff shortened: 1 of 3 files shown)


NeroOneTrueKing (approved):
Didn't pull to test, but reading through it looks good!

PR #1521, opened 2021-05-09
adds seven-league-boots

Tarmunora:
unaligned artifact jumping boots:
plussev, extrinsic speed
6x speed for multi-tile-movement
invoke for 7-tile jump

more artifacts... good?

### DIFF
--- a/src/allmain.c
+++ b/src/allmain.c
@@ -931,6 +931,8 @@ moveloop()
 				youmonst.movement -= 3;
 			} else if(uwep && uwep->oartifact == ART_TOBIUME){
 				youmonst.movement -= 4;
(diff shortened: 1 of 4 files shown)


src/allmain.c
```diff
@@ -931,6 +931,8 @@ moveloop()
 				youmonst.movement -= 3;
 			} else if(uwep && uwep->oartifact == ART_TOBIUME){
 				youmonst.movement -= 4;
+			} else if(uarmf && uarmf->oartifact == ART_SEVEN_LEAGUE_BOOTS){
```
NeroOneTrueKing: This should be earlier in the else-if chain. You shouldn't slow down if wielding Tobiume.

NeroOneTrueKing (changes requested):

NeroOneTrueKing (approved):

PR #1559, opened 2021-06-07
Tobiume now tases enemies

RikerW:
Tobiume still deals +1 fire damage, but the on-death explosions and invoke now deal a combination of fire and shock explosions. It also gives both fire and shock resistance when wielded, up from just fire.

If the unskilled invoke is used (single beam), the damage type is chosen randomly. If the skilled invoke is used (cluster of explosions), the damage type is chosen randomly and independently for each explosion.

For the on-death explosions:
* If the monster resists fire but not shock, it will use a shocking explosion
* If the monster resists shock but not fire, it will use a fiery explosion
* If the monster resists neither fire nor shock, it will use either one with a 50% chance each
* If the monster resists both, it will use a fiery explosion

Also fixes a minor missing period if the player does not have enough energy to use the shikai.

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -5019,7 +5019,7 @@ arti_invoke(obj)
     if(oart->inv_prop > LAST_PROP) {
 	/* It's a special power, not "just" a property */
 	if(obj->age > monstermoves && 
-		oart->inv_prop != FIRE_SHIKAI && 
+		oart->inv_prop != TOBIUME_SHIKAI &&
 		oart->inv_prop != SEVENFOLD && 
 		oart->inv_prop != ANNUL && 
 		oart->inv_prop != ALTMODE && 
@@ -5041,7 +5041,7 @@ arti_invoke(obj)
 	    return partial_action();
 	}
 	if( /* some properties can be used as often as desired, or track cooldowns in a different way */
-		oart->inv_prop != FIRE_SHIKAI &&
+		oart->inv_prop != TOBIUME_SHIKAI &&
 		oart->inv_prop != ICE_SHIKAI &&
 		oart->inv_prop != NECRONOMICON &&
 		oart->inv_prop != SPIRITNAMES &&
@@ -5675,7 +5675,7 @@ arti_invoke(obj)
 			}
 		}
 	break;
- 	case FIRE_SHIKAI:
+	case TOBIUME_SHIKAI:
 		{
 			boolean toosoon = monstermoves < obj->age;
 			if ( !(uwep && uwep == obj) ) {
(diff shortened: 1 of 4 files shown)


src/xhity.c
```diff
@@ -13977,6 +13977,11 @@ int vis;						/* True if action is at all visible to the player */
 			elemdmg += 6 + weapon->spe;
 			totldmg += 6 + weapon->spe;
 		}
+	} else if(weapon && weapon->oartifact == ART_TOBIUME && !Shock_res(mdef)){
```
NeroOneTrueKing, line 13980: This `else if` could just be an `||` in the original `if`, since the bodies of the two are identical.
`if(weapon && weapon->oartifact == ART_TOBIUME && !(Fire_res(mdef) || Shock_res(mdef))`

src/artifact.c
```diff
@@ -5687,15 +5687,13 @@ arti_invoke(obj)
 				int role_skill;
 				struct obj *pseudo;
 				coord cc;
+				pseudo = mksobj(SPE_FIREBALL, MKOBJ_NOINIT); // create here for the skill check
```
NeroOneTrueKing, line 5690: (Just commenting on the first time it happens)
`mksobj()` is being called, but the object is making is never being deleted. Yeah the original Tobiume was bad about this too, but could you clean this up while you're here? I think it just needs a `delobj()` at the end after each one is used. It may also be convenient to call `mksobj()` less.

src/artifact.c
```diff
@@ -5687,15 +5687,13 @@ arti_invoke(obj)
...
 
 				energy = toosoon ? 25 : 15;
-				pseudo = mksobj(SPE_FIREBALL, MKOBJ_NOINIT);
-				pseudo->blessed = pseudo->cursed = 0;
-				pseudo->quan = 20L;			/* do not let useup get it */
 				role_skill = max(P_SKILL(uwep_skill_type()), P_SKILL(spell_skilltype(pseudo->otyp)) );
```
NeroOneTrueKing, line 5693: For example (about calling mksobj less), why not replace `pseudo->otyp` with `SPE_FIREBALL`?

NeroOneTrueKing (changes requested):
Just a few cleaner-code things, if you'd please :)

Chris-plus-alphanumericgibberish:
I think Tobiume is pretty clearly fiery rather than electrical in Bleach.

Perhaps the fire explosions should be set to Elemental (deal half damage to resistant targets) instead.

Resistances need a re-balance :(

RikerW:
I made this at 3 am, this was stupid then and it's equally stupid now. Fun to mess around with in a joke (ndnh assblaster???) run though. Closing.

PR #1560, opened 2021-06-07
Add quick check-and-abort to priest_talk() if mon isn't a real priest

NeroOneTrueKing:
Notably, The Stranger currently has the MS_PRIEST sound, which causes a crash if you try to talk to it.

### DIFF
--- a/src/priest.c
+++ b/src/priest.c
@@ -558,6 +558,11 @@ void
 priest_talk(priest)
 register struct monst *priest;
 {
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
The fact that the Stranger is a priest is a copy-paste error, but this is good futureproofing regardless.

Don't really know what the Stranger **should** be.

PR #1565, opened 2021-06-19
Modfiy DEADMONSTER() macro to claim player is alive, not dead.

### LINKED ISSUE #1564: DEADMONSTER should include "and the monster is not the PC"
The player character doesn't use the youmonst hp field, so they always register as dead and need to be excluded from the DEADMONSTER check.

Marked "complex" because removing the existing != youmonst checks sounds like a pain in the neck.

NeroOneTrueKing:
We can avoid checking the player's hp, as when the player is truly dead, things tend to be quickly resolved anyways.

### DIFF
--- a/src/xhity.c
+++ b/src/xhity.c
@@ -812,7 +812,7 @@ int tary;
 													(dx || dy) ? m_at(tarx + dx, tary + dy) : 
 													(struct monst *)0;
 							if (mdef2 
-								&& (!DEADMONSTER(mdef2) || mdef2 == &youmonst)
+								&& (!DEADMONSTER(mdef2))
 								&& ((!youagr && mdef2 != &youmonst && mdef2->mpeaceful != magr->mpeaceful) ||
 									(!youagr && mdef2 == &youmonst && !magr->mpeaceful) ||
 									(youagr && !mdef2->mpeaceful))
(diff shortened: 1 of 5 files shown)


src/xhity.c
```diff
@@ -745,9 +744,7 @@ int tary;
 							u.uy == tary - dy
 							)
 						{
-							struct monst *mdef2 = u.uswallow ? u.ustuck :
```
Chris-plus-alphanumericgibberish: There's a regression here. This specific instance won't lead to the player attacking themselves, but it's unlikely they ever want to attack their own square.

src/xhity.c
```diff
@@ -807,12 +802,9 @@ int tary;
 							y(magr) == tary - dy
 							)
 						{
-							struct monst *mdef2 = !youagr ? m_u_at(tarx + dx, tary + dy) :
```
Chris-plus-alphanumericgibberish: This regresses the engulfer-self-attack bug :(

Chris-plus-alphanumericgibberish (dismissed):

NeroOneTrueKing:
Not sure how that regression snuck in, so I remade the branch and force-pushed. All better now!

Chris-plus-alphanumericgibberish (approved):

PR #1616, opened 2021-08-02
Player Androids don't inherit ShockRes from their race

### LINKED ISSUE #1581: Androids are getting intrinsics from their base monster type`
Start with shock res, and shouldn't.

NeroOneTrueKing:
They get it from race only when leveling up.

Unfortunately, this adds ugliness to youprop.h; oh well.


Chris-plus-alphanumericgibberish (dismissed):

Chris-plus-alphanumericgibberish (changes requested):
This also affects androids and fire resistance.

I was also expecting it to affect clockworks, but it seems that isn't a problem.

Chris-plus-alphanumericgibberish (approved):

PR #1641, opened 2021-08-12
Exempt ball&chain from being scattered

### LINKED ISSUE #1622: Scatter() fails to handle the ball and chain
The easiest way to recreate this is to wish up a landmine and a source of flying, put the ball and chain on one square, set the mine there, and then fly back and forth until the mine triggers.

If the PC falls into the pit, the ball and chain are pulled in after them, hiding the bug.

Does not appear to cause a crash, so there's that.

NeroOneTrueKing:


Chris-plus-alphanumericgibberish:
I think this should be resolved the other way: if a player can arrange to blow up their ball and chain, they should become unpunished.

The exploding chest code already does this, at least to some extent, but land mines don't.

Chris-plus-alphanumericgibberish (approved):

PR #1666, opened 2021-08-19
Remove accidental code duplication, move all guaranteed-failures to top

NeroOneTrueKing:
Just cleanup: things that prevent spellcasting even with Naen active go to the very top.*

### DIFF
--- a/src/spell.c
+++ b/src/spell.c
@@ -5665,6 +5665,9 @@ int spell;
 		return 0;
 	if(mad_turn(MAD_TOO_BIG))
 		return 0;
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (dismissed):

Chris-plus-alphanumericgibberish:
This duplication is required rather than accidental. Re-casts of maintained spells and non-menu-based spellcasting go straight to spell failure without passing through the menu code.

NeroOneTrueKing:
It's reduplicated within the same function (in addition to the spell menu code noping out on its own):
image

PR #1721, opened 2021-09-09
Avoid having containers as vault loot

NeroOneTrueKing:
Logic was backwards, prefering containers.


Chris-plus-alphanumericgibberish:
This is not an accident. Containers may spawn with contents, and contents aren't handled with any degree of care in this code :-/. In particular, I'm worried that a container could spawn with an artifact inside, which would then be removed from the game. I haven't checked the container code to see if contents can be artifacts, though.

NeroOneTrueKing:
Ah, I see

PR #1763, opened 2021-09-26
Make overcharged blasters blow the heck up

RojjaCebolla:


NeroOneTrueKing (commented):
Nice! I feel like the message should be printed before the explosion happens, though.

RojjaCebolla:
Cool, I'll change that.

Riker helped me build DNH correctly, so it turns out the game doesn't even allow *attempting* to charge a blaster for the 5th time. Which maybe means that "if >= 4" check was probably never actually needed/checked?

I'll keep hunting for the way to make it happen, unless you know what the prob might be.

RojjaCebolla:
oh, nevermind, it works! i just needed to **start a fresh game** after recompiling

RojjaCebolla:
yay, finally got rid of the dup commits from upstream. don't git pull in a PR branch, folks!

RojjaCebolla:
actually i'm not done with this, it should warn you once before the gun blows up

RojjaCebolla:
ok, all done!

NeroOneTrueKing (approved):

Chris-plus-alphanumericgibberish:
I'm closing this version because it's against 3.21.1

Copypasta from the issue:

I mentioned this in chat a while ago, but the issue with https://github.com/Chris-plus-alphanumericgibberish/dNAO/pull/1763 is that blaster charge limits are referencing a second ammo type rather than overheating.

However, I'm still on board with incorporating it in some form. For instance, perhaps charging an already-charged blaster might cause it to explode.

PR #1809, opened 2021-11-03
Fix a couple cases when artifacts could get deleted

RikerW:
Various silent artifact deletions plus two cases where I think artifact chests should be marginally more durable (aka Proteus I guess). Credit to entrez for commits 2abe08 and b242fb on vanilla 3.7 for the silent ones.

### DIFF
--- a/src/shk.c
+++ b/src/shk.c
@@ -1004,6 +1004,7 @@ register struct obj *obj;
 
 	while ((curr = obj->cobj) != 0) {
 	    obj_extract_self(curr);
+	    if (curr->oartifact) artifact_exists(curr, ONAME(curr), FALSE);
 	    obfree(curr, (struct obj *)0);
 	}
 }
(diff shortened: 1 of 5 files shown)


src/shk.c
```diff
@@ -1004,6 +1004,7 @@ register struct obj *obj;
 
 	while ((curr = obj->cobj) != 0) {
 	    obj_extract_self(curr);
+	    if (curr->oartifact) artifact_exists(curr, ONAME(curr), FALSE);
```
NeroOneTrueKing, line 1007: `delete_contents()` should not set artifacts as not being created -- polymorphing a container calls this, as does exploding a bag of holding. It would then be possible to put an artifact inside a container, destroy it, and then get the artifact again.

For bonus points, put a single Fluorite Octahedron in a box, then poly the box, then wish for another Octet and watch the game break.

NeroOneTrueKing (changes requested):

RikerW:
I'll remake this eventually to the right branch.

PR #1825, opened 2021-12-31
Don't make doors in not okay places

NeroOneTrueKing:
Typically, the not-okay place to make a door is on top of another door so it wasn't noticed, except it consumed precious `doors[]` indices, thus rarely causing `DOORMAX` to be hit on dense levels (like Mithardir catacombs).

### DIFF
--- a/src/mklev.c
+++ b/src/mklev.c
@@ -327,7 +327,7 @@ boolean nxcor;
 			return;
 
 		// add the door
-		if (okdoor(xx, yy) || !nxcor)
+		if (okdoor(xx, yy))
 		{
 			dodoor(xx, yy, croom);
 			add_door(xx, yy, troom);
@@ -342,7 +342,7 @@ boolean nxcor;
 	if (IS_WALL(levl[xx + dx][yy + dy].typ) || IS_ROOM(levl[xx + dx][yy + dy].typ))	// prevent trying to open corridors into adjacent rooms
 		return;
 
-	if (okdoor(xx,yy) || !nxcor)
+	if (okdoor(xx,yy))
 	    dodoor(xx,yy,croom);
 
 	org.x  = xx+dx; org.y  = yy+dy;
@@ -353,7 +353,7 @@ boolean nxcor;
 	    return;
 
 	/* we succeeded in digging the corridor */
-	if (okdoor(tt.x, tt.y) || !nxcor)
+	if (okdoor(tt.x, tt.y))
 	    dodoor(tt.x, tt.y, troom);
 
 	if(smeq[a] < smeq[b])


Chris-plus-alphanumericgibberish (commented):
I'm confused why this is the solution to the bug. Is this a vanilla error that was never detected? Do you know what nxcor is doing there? It seems like an oddly named variable and a cursory read didn't clue me in to what it's trying to do.

NeroOneTrueKing:
Breaking down `join()` (comments added in place)
```c
	croom = &rooms[a]; // current room
	troom = &rooms[b]; // target room
```
The next block picks based on the relative locations of current and target rooms. I'll just pull apart the first `if`:
```c
	if(troom->lx > croom->hx) { // target room is fully to the right of current room
		dx = 1; // corridor traveling to the right
		dy = 0; // non vertical
		xx = croom->hx+1; // column on the right wall of current room
		tx = troom->lx-1; // column on the left wall of target room
		finddpos(&cc, xx, croom->ly, xx, croom->hy); // find a position 'cc' in the right wall of current room
		finddpos(&tt, tx, troom->ly, tx, troom->hy); // find a position 'tt' in the left wall of target room
	}
```
Then we set variables
```c
	xx = cc.x; // door position on current room
	yy = cc.y;
	tx = tt.x - dx; // square directly in front of wall on target room
	ty = tt.y - dy;
```
And we get our first instance of `nxcor`!
```c
	if(nxcor && levl[xx+dx][yy+dy].typ) // if the square directly in front of the wall in our current room is anything but stone
		return; // end early, join() did nothing
```
I'll skip the shared-wall-door-making, as that's not vanilla.
Our second use of `nxcor`...
```c
	// on wall of current room
	if (okdoor(xx,yy) || !nxcor) // if it's an ok place for a door, or !nxcor
	    dodoor(xx,yy,croom); // make the door here
```
Dig a corridor, return early if it fails... `nxcor` here allows the corridor to fail 'for no reason' partway through.
```c
	org.x  = xx+dx; org.y  = yy+dy;
	dest.x = tx; dest.y = ty;

	if (!dig_corridor(&org, &dest, nxcor,
			level.flags.arboreal ? ROOM : CORR, STONE))
	    return;
```
And our final use of nxcor:
```c
	// on wall of target room
	if (okdoor(tt.x, tt.y) || !nxcor) // if it's an ok place for a door, or !nxcor
	    dodoor(tt.x, tt.y, troom); // make the door here
```
Huh.

Now where can these target locations be? [xx,yy] was determined earlier by `finddpos()`:
```c
finddpos(cc, xl,yl,xh,yh)
coord *cc;
xchar xl,yl,xh,yh;
{
	register xchar x, y;

	x = (xl == xh) ? xl : (xl + rn2(xh-xl+1));
	y = (yl == yh) ? yl : (yl + rn2(yh-yl+1));
	if(okdoor(x, y))
		goto gotit;

	for(x = xl; x <= xh; x++) for(y = yl; y <= yh; y++)
		if(okdoor(x, y))
			goto gotit;

	for(x = xl; x <= xh; x++) for(y = yl; y <= yh; y++)
		if(IS_DOOR(levl[x][y].typ) || levl[x][y].typ == SDOOR)
			goto gotit;
	/* cannot find something reasonable -- strange */
	x = xl;
	y = yh;
gotit:
	cc->x = x;
	cc->y = y;
	return;
}
```
Which can ONLY get coordinates of `okdoor()` or `DOOR` or `SDOOR` (or the bottom left corner, which we don't see happening).
Therefore, all that the `!nxcor` is doing is allowing us to add doors on top of doors. Which is dumb. Vanilla doesn't run into the maxdoors issue because I amped up roomcount in dnh levelgen, so it makes many more attempts to `join()` rooms and can create doors on top of doors like this enough times to reach `MAXDOOR`.

Chris-plus-alphanumericgibberish (approved):

PR #1831, opened 2022-01-06
All sickles are throwing weapons

NeroOneTrueKing:
Only Sickle Moon is returning.


Chris-plus-alphanumericgibberish (approved):

Chris-plus-alphanumericgibberish:
I'm going to comment out the code that lets monsters throw sickles: makemon provides several monsters with sickles as their primary weapon and the throw code doesn't check if the monster is throwing away their melee weapon :(

PR #1841, opened 2022-01-08
Impossible() on bad explosion locations

NeroOneTrueKing:


NeroOneTrueKing:
On second thought this isn't necessary. `isok()` is checked prior to creating the explode area.

PR #1857, opened 2022-01-12
Skip des coordinate sanity checks

amateurhourrrr:
This allows an optional flag, [skipcheck] or [noskipcheck]
to prevent / force the coordinate check routine.

Presently it's only implemented for objects, monsters, and engravings.

If this ever needs extending with more categories, it would be worthwhile
to instead extend the coord data type and track a skip flag there,
but I wanted to keep it pretty simple here


NeroOneTrueKing (approved):
Looks good, thanks!

PR #1887, opened 2022-03-13
Make `MM_MISS` return mean no time taken, not miss.

NeroOneTrueKing:
Fix pets doing nothing if they try to attack something and their attack chain does utterly nothing (incl. messages, like wild misses).

Would need to rewrite monster movement to be able to fix this for non-tame mon v mon combat -- they pick the square to moveto/attack too strongly at the moment to retarget.

### DIFF
--- a/src/dogmove.c
+++ b/src/dogmove.c
@@ -1094,7 +1094,7 @@ register int after;	/* this is extra fast monster movement */
 				if (res & MM_AGR_DIED)
 					return 2; /* Oops, died */
 
-				if (!(mon_ranged_gazeonly) && (res & MM_HIT))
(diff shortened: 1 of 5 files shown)


Chris-plus-alphanumericgibberish (dismissed):

Chris-plus-alphanumericgibberish:
Doesn't this alter the behavior of passive attacks?

Example, near xhity 16030 ("These passives happen even if the defender died, but require a hit")

NeroOneTrueKing:
Aaaaaaaaa yes it would, I need to review this

NeroOneTrueKing:
Closing PR as-is. It would be inconsistent to make MM_MISS sometimes mean "no time taken" and sometimes mean "no attack hit". MvM behaviour being weird is nothing new.

PR #1901, opened 2022-03-22
Modify objclass property defn to be array

NeroOneTrueKing:
List all properties inside `{ }`, or in `O_POWER()`.

Some things (esp do_wear.c) still only check the 1st property.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -3616,8 +3616,11 @@ winid *datawin;
 		boolean got_prop = FALSE, while_carried = FALSE;
 		int j = 0;
 
-		if (oc.oc_oprop == propertynames[i].prop_num)
-			got_prop = TRUE;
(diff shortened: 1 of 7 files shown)


src/invent.c
```diff
@@ -3616,8 +3616,11 @@ winid *datawin;
 		boolean got_prop = FALSE, while_carried = FALSE;
 		int j = 0;
 
-		if (oc.oc_oprop == propertynames[i].prop_num)
-			got_prop = TRUE;
+		while(oc.oc_oprop[j] && !got_prop) {
```
Chris-plus-alphanumericgibberish, line 3619: Shouldn't this have && j < OC_PROP_MAX or similar?

Chris-plus-alphanumericgibberish (approved):

PR #1927, opened 2022-04-23
Variable-speed single actions

NeroOneTrueKing:
This takes what was previously being done with moving (reduced movement point cost under certain conditions) and generalizes it for attacking and future expansion into other action types.

Attack actions now cost less (or more) movement points at the place where movement points are deducted, instead of giving/costing movement points in `attack2()` (note: that function can be renamed back to `attack()` now).

Shows action type and time taken in status bar, for wizardmode in curses only, for my debugging.

TODO: create more `MOVE_xyz` flags and work my way through the various functions that currently return `1` for time taken and `0` for instant, to be able to do things like "this artifact makes drinking potions take less time" etc.


src/cmd.c
```diff
@@ -2914,17 +2914,21 @@ register char *cmd;
...
+
+			/* backwards compatibility with functions returning 0 to mean no time taken */
+			if (res==0)
+				res = MOVE_INSTANT;
 		}
-		if (!res) {
-		    flags.move = FALSE;
-		    multi = 0;
+		flags.move = res;
+		if (res == MOVE_CANCELLED || res == MOVE_INSTANT) {
```
Chris-plus-alphanumericgibberish: Should this be res&MOVE_CANCELLED ?

NeroOneTrueKing:
Oh shit git time >_>

Chris-plus-alphanumericgibberish (approved):

src/cmd.c
Chris-plus-alphanumericgibberish: This is old

PR #1935, opened 2022-04-30
Remove excessive color desc from star of hypernotus

NeroOneTrueKing:
The 'star sapphire' material already acts as a color, and explicitly so when unIDed.

### DIFF
--- a/include/artilist.h
+++ b/include/artilist.h
@@ -2100,7 +2100,7 @@ A("Idol of Bokrug, the water-lizard",		STATUE,	"sea-green stone likeness of a wa
 	NOINVOKE, NOFLAG
 	),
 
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish (commented):
The elder sign being greenish instead of blue is important (look up "fairy coin fossil" for examples), so I think the better resolution is to squelch the material if star-saphire hasn't been identified (the Star is playing *really* loose with materials)

RikerW:
Wouldn't it be easier to just have the material code have an exception to describe the Star of Hypernotus with a different color? Similar to the handling for MINERAL -> enamel for crysknives & worm teeth, or onyx for the Lash of the Cold Waste. 

Edit - PR made by me with a slightly different fix.

PR #1979, opened 2022-07-10
Add guards to loss in scatter

### LINKED ISSUE #1972: Eggs falling out of a tree and hitting you causes a crash.
This is possible in pirate quest , drow quest, and dwarf noble quest. also paradise island.
 
image

amateurhourrrr:
It is intentional that loss can be null in common circumstances, but in
the case of e.g. eggs here there weren't any guards for it, causing a
segfault

fixes #1972


NeroOneTrueKing (approved):

demogorgon22:
sounds more like an eggfault than a segfault... :)

PR #2041, opened 2022-09-27
enlighten.c: terminal punctuation in attribute display

NetSysFire:
This does not apply to the dumplog it appears, since it uses `enl_msg`. Seeing the missing terminal punctuation in the attribute display (ctrl+x) has annoyed me to no end, but now it should be gone for good.
Includes bonus minor typo fixes.

### DIFF
--- a/src/enlighten.c
+++ b/src/enlighten.c
@@ -1400,44 +1400,44 @@ resistances_enlightenment()
 	}
 	
 	if(u.wimage >= 10){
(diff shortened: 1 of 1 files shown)


NeroOneTrueKing (approved):

NetSysFire:
image

This is basically the only thing I worry about. I don't know if the additional terminal punctuation will wrap the sentence around.

PR #2045, opened 2022-10-10
Make more levels bones eligible!

NeroOneTrueKing:
Portals and Branches no longer make levels bones-ineligible.

Portals
* The level branching to the Quest is explicitly ineligible, because that portal can be destroyed.
* Branches accessed via portal try to find the portal on the bones level leading to the other end of the branch.

Branches
* Bones save branch levels with an ID indicating the Dungeon that was being branched to (or none)
* * ex, ID is "MainDungeon.NeutralEntrance.BranchingToNeutrality"
* * ex, ID is "MainDungeon.dLevel16.BranchingToLostTomb"
* * ex, ID is "LostTomb.LostTomb.BranchingToMainDungeon"
* Bones don't save exact level (Special char / dlevel) because it makes dlevel branches much less likely to be found, and there are no side effects that I can think of.

### DIFF
--- a/dat/dngnch1.def
+++ b/dat/dngnch1.def
@@ -37,13 +37,13 @@ LEVALIGN:	chaotic
 RNDLEVEL:	"castle" "none" @ (-1, 0) 6
 CHAINBRANCH:	"Gehennom" "castle" + (0, 0) no_down
 
-RNDLEVEL:		"lawlev" "none" @ (12, 8) 3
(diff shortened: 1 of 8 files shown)


NeroOneTrueKing:
I haven't checked to be sure, but this should be both save compatible and existing-bones compatible.

Chris-plus-alphanumericgibberish (approved):

PR #2055, opened 2022-10-14
Add rcfile option to set fg color, bg color, and symbol for any template

### LINKED ISSUE #2039: Pseudonatural templated enemies should be highlighted
As it stands, they can randomly generate and look no different than regular monsters while being among the most dangerous monsters. They should be highlighted some color like zombies are so they stand out right away and the player can know "oh that's a brain eater that can kill me through my life saving" and not just think it is another normal monster. As it stands, it is very antiqol needing to manually check each monster from a range to make sure it doesn't have that very deadly template.

NeroOneTrueKing:
Syntax:
MONSTERTEMPLATE = <template name> : <fg / bg / symbol> : <color/character>

ex
```
MONSTERTEMPLATE = zombified     :symbol :Z
MONSTERTEMPLATE = zombie        :bg     :green
MONSTERTEMPLATE = illuminated   : bg    :yellow
MONSTERTEMPLATE = tomb herd     :fg     :gray
MONSTERTEMPLATE = vitrified     :background : cyan
MONSTERTEMPLATE = pseudonatural :foreground : lightmagenta
```

Note: on TTY, it only has 8 colors of background available (no bright colors)

Does not *break* saves, though games in progress will have hallucination-like rememberances of areas explored prior to the update because the glyphs are shifted. These can be somewhat fixed by ctrl-r to redraw terrain, and then exploring to re-see the tiles that ctrl-r doesn't fix.

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -321,8 +321,8 @@
 #define GLYPH_MON_OFF		0
 #define GLYPH_PET_OFF		(NUMMONS	+ GLYPH_MON_OFF)
 #define GLYPH_PEACE_OFF		(NUMMONS	+ GLYPH_PET_OFF)
-#define GLYPH_ZOMBIE_OFF	(NUMMONS	+ GLYPH_PEACE_OFF)
-#define GLYPH_INVIS_OFF		(NUMMONS	+ GLYPH_ZOMBIE_OFF)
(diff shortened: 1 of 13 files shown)


Chris-plus-alphanumericgibberish:
Should we put this on devel instead of compat?

Chris-plus-alphanumericgibberish (approved):

PR #2072, opened 2022-10-23
Fix blocking in plane of earth (I think)

### LINKED ISSUE #2065: Earth plane ceiling collapsing doesn't reblock previously dug out squares
This allows monsters to gaze attack you through the stone, you to see through the stone as if it isn't there, and for monsters to hit you with reach attacks through the stone.

NeroOneTrueKing:
Closes #2065

### DIFF
--- a/src/dig.c
+++ b/src/dig.c
@@ -85,6 +85,11 @@ mkcavepos(x, y, dist, waslit, rockit)
     if(Blind)
 	feel_location(x, y);
     else newsym(x,y);
+
+	if (rockit)
+		block_point(x, y);
+	else
+		unblock_point(x, y);
 }
 
 STATIC_OVL void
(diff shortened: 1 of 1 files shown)


src/dig.c
```diff
@@ -85,6 +85,11 @@ mkcavepos(x, y, dist, waslit, rockit)
     if(Blind)
 	feel_location(x, y);
     else newsym(x,y);
+
+	if (rockit)
+		block_point(x, y);
+	else
+		unblock_point(x, y);
```
Chris-plus-alphanumericgibberish, line 92: I think this needs to check if(!does_block()). There might hypothetically be a boulder on that spot.

src/dig.c
```diff
@@ -2284,13 +2289,17 @@ int oldy;
...
 			levl[tx][ty].typ = STONE;
+			block_point(tx, ty);
+		}
 	/* dig out around new position */
 	for (tx = max(0, mtmp->mx-1); tx < min(COLNO, mtmp->mx+1); tx++)
 	for (ty = max(0, mtmp->my-1); ty < min(ROWNO, mtmp->my+1); ty++)
-		if (levl[tx][ty].typ == STONE && may_dig(tx, ty))
+		if (levl[tx][ty].typ == STONE && may_dig(tx, ty)) {
 			levl[tx][ty].typ = CORR;
+			unblock_point(tx, ty);
```
Chris-plus-alphanumericgibberish, line 2301: This may also need if(!does_block())

Chris-plus-alphanumericgibberish (commented):
I think I did the commenting wrong :(

Chris-plus-alphanumericgibberish (approved):

PR #2126, opened 2023-05-07
add fallingstar mandibles (new artifact)

RikerW:
astel's head on a stick
pincer staff that's dubiously alive and shoots explosions, rocks and lasers +12 to hit, +d12 magic damage
rated at B because fuck it that's why
goes to +10
invoke to throw some magical explosions at things adjacent or around you, dealing 3d12+3xspe. invoke up triggers living effects also does 2x base to is_aliens, an extra 1d12 magic on grabbed targets above 20 insight as living when wielded, above 40 insight is alive, randomly (insight > d100) throws 4d(spe/2) rocks at non-flying targets within 3 tiles occasionally as an invoke, can end rock attacks with a 6d6 magic beam towards that target, doesn't bounce/can't reflect. only possible when invoked

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -9857,7 +9857,7 @@ arti_invoke(obj)
 				obj->age = monstermoves;
 				return MOVE_CANCELLED;
 			}
-#define astel_blast(x, y, size) explode(x, y, AD_MAGM, WAND_CLASS, (d(1, 12) + uwep->spe) * ((Double_spell_size) ? 3 : 2) / 2, EXPL_CYAN, size + !!Double_spell_size)
+#define astel_blast(x, y, size) if (isok(x, y)) explode(x, y, AD_MAGM, WAND_CLASS, (d(1, 12) + uwep->spe) * ((Double_spell_size) ? 3 : 2) / 2, EXPL_CYAN, size + !!Double_spell_size)
 			if (!getdir((char *)0)) { //Oh, getdir must set the .d_ variables below.
 			    /* getdir cancelled, just do the nondirectional scroll */
 				obj->age = 0;
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -4699,8 +4699,12 @@ int dieroll; /* needed for Magicbane and vorpal blades */
 	}
 
 	if(otmp->otyp == PINCER_STAFF && u.uinsight >= 10){
-		if(otmp->ovar1_pincerTarget == mdef->m_id)
+		if(otmp->ovar1_pincerTarget == mdef->m_id){
 			*plusdmgptr += basedmg;
+			if (u.uinsight >= 20 && otmp->oartifact && otmp->oartifact == ART_FALLINGSTAR_MANDIBLES && !Magic_res(mdef)){
+				*truedmgptr += reduce_dmg(mdef, d(1, 12), FALSE, TRUE);
+			}
+		}
 		else otmp->ovar1_pincerTarget = mdef->m_id;
 	}
 }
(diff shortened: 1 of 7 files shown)


src/artifact.c
```diff
@@ -4699,8 +4699,12 @@ int dieroll; /* needed for Magicbane and vorpal blades */
 	}
 
 	if(otmp->otyp == PINCER_STAFF && u.uinsight >= 10){
-		if(otmp->ovar1_pincerTarget == mdef->m_id)
+		if(otmp->ovar1_pincerTarget == mdef->m_id){
 			*plusdmgptr += basedmg;
+			if (u.uinsight >= 20 && otmp->oartifact && otmp->oartifact == ART_FALLINGSTAR_MANDIBLES && !Magic_res(mdef)){
+				*truedmgptr += reduce_dmg(mdef, d(1, 12), FALSE, TRUE);
```
demogorgon22, line 4705: Does it make sense to call reduce_dmg for this dmg? What does reduce_dmg do? I don't know, I'm just asking.

src/artifact.c
```diff
@@ -5425,6 +5429,13 @@ boolean printmessages; /* print generic elemental damage messages */
 		if (Hallucination) You("flame the nasty troll!");  //trollsbane hits monsters that pop in to ruin your day.
 		*truedmgptr += d(2, 20) + 2 * otmp->spe; //boosts power better than demonbane hitting silver hating.
 	}
+	
+	if (youagr && is_alien(pd) && oartifact == ART_FALLINGSTAR_MANDIBLES)
+	{
+		*truedmgptr += basedmg;
```
demogorgon22, line 5435: Is this a pattern any other artis follow? Is this damage described in the inspection menu for the weapon?

src/artifact.c
```diff
@@ -9839,6 +9850,47 @@ arti_invoke(obj)
...
 			}
 
+		break;
+		case WAVES_DARKNESS:
+			if (!uwep || uwep != obj){
+				You_feel("that you should be wielding %s.", the(xname(obj)));
+				obj->age = monstermoves;
+				return MOVE_CANCELLED;
+			}
+#define astel_blast(x, y, size) explode(x, y, AD_MAGM, WAND_CLASS, (d(1, 12) + uwep->spe) * ((Double_spell_size) ? 3 : 2) / 2, EXPL_CYAN, size + !!Double_spell_size)
```
demogorgon22: Add an isok check to x and y before calling this as the code below can potentially call it off the map and I think calling explode off of the map is still BAD.

src/artifact.c
demogorgon22: I assume this is for checking half spell dam maybe?

demogorgon22 (commented):
Thanks, needed to find the es objnam.c zone last night and was looking for this PR but couldn't find it. Added some comments, not exhaustive, just some thoughts.

src/artifact.c
RikerW: yeah, it's for half spell/phys dam (spell here). see above lines about like, diskos extra damage & such. it uses truedmgptr to bypass any spec_applies check, since it checks Magic_res already.

src/artifact.c
RikerW: yeah, this is the section for trollsbane extra damage, giantslayer slow, plague explosions, etc. it does NOT show up in arti inspect unfortunately, but I could/will probably add a manual line to that display that says as much. trollsbane et. al. should probably get that line added too though :/

src/artifact.c
RikerW: 👍

src/artifact.c
RikerW: added a line in invent.c to reflect that, see new commit

src/do_wear.c
```diff
@@ -5439,6 +5439,80 @@ boolean invoked;
...
+					return; //oops!
+			}
+			
+			if(u.uinsight >= 50 && invoked && !rn2(4)){
+				pline("A torrent of energy erupts from the jaws!");
+				struct zapdata beam;
+				basiczap(&beam, AD_MAGM, ZAP_BREATH, 6);
+				beam.unreflectable = ZAP_REFL_NEVER;
+				beam.no_bounce = TRUE;
+				zap(magr, x(magr), y(magr), sgn(x(mdef) - x(magr)), sgn(y(mdef) - y(magr)), BOLT_LIM, &beam);
```
Chris-plus-alphanumericgibberish, line 5510: Doesn't this target all creatures in a 7x7 square, potentially resulting in weirdness with beam angles?

Also, can't this blast peacefuls and pets that are lined up with the primary target, getting the PC blamed for murder?

Chris-plus-alphanumericgibberish (changes requested):

Chris-plus-alphanumericgibberish (approved):

PR #2152, opened 2023-07-09
Allow the Oracle to tell you about Demon/Devil Lords

RikerW:
Reworks how mkmaze.c handles picking a level for abyss, abs2, brine, hell1, hell2. Does NOT affect any other special levels or actually how it generates any, just forces the appropriat level var to match the dungeon topology variant listing instead of picking a random level var and setting the variant listing to match. Should have unchanged behavior from before. The one change I actually made about the levels was to give the chromatic dragon an actual flag for herself (`CHROMA_LEVEL`), with `Is_chromatic_level` handling the checks. However, this is pretty useless, and only currently affects the level name in #overview (now "Dragon Caves" instead of "Upper Hell" for her).

Oracle hints for each thing are more than open to suggestions - seriously, I'm writing all of these in one go and I am NOT a poet by any means. Some phrasings could be less awkward and some messages more inspired.

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -83,6 +83,7 @@ E struct dgn_topology {		/* special dungeon levels for speed */
 #define DISPATER_LEVEL	2
 #define MAMMON_LEVEL	3
 #define BELIAL_LEVEL	4
+#define CHROMA_LEVEL	5
 	d_level	d_hell2_level;
 	int		hell2_variant;
 #define LEVIATHAN_LEVEL	1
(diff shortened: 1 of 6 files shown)


RikerW:
Another possibility is adjusting price or access - right now hints are sitting at $150 for any hint. This is pretty reasonable for the other hints, since most of them are easy enough to find out (chaos variant, oona energy, plat dragon) or mostly useless (tulani/etc. type, polypoid being) but I didn't bother changing the price of these new hints specifically until I had suggestions on what to set it to. I think around $1000 sounds reasonable, but maybe also just $500 and limit it to XP 10+ or veil gone? Unsure.

note - second PR since i fucked up and made the first one to compat and rebasing sucks and it was easier to just make a new PR

Chris-plus-alphanumericgibberish (approved):

PR #2158, opened 2023-07-23
New artifact: Enforced Mind

RikerW:
* lawful helm of telepathy (default material)
* does not actually grant telepathy. instead, will either strengthen your telepathy (if you have telepathy from some other source) or entirely block telepathy (if you don't have it from some other source)
* if it strengthens your telepathy, will boost intrinsic to extrinsic, and make the extrinsic range stretch 4x as far as the normal range. in addition, any monsters that you can see with extrinsic telepathy (so not cross-level if you're blind) are treated as if they have telepathy of their own. this grants them all appropriate bonuses or vulnerabilities to identifying you, eating mind blasts, etc.
* if it blocks your telepathy, makes you immune to being seen by telepaths (so they don't know your location from telepathy at least), and nullifies any incoming sources of psionic bonus damage or mind blasts. ** todo: this does not check monster spells. there are probably some that should be blocked, esp for psi casters
* boosts charisma by it's enchantment when worn, grants a charisma bonus (same rules as dexterity bonus and is treated as such re armor etc.) to AC
* grants confusion+stun res (like hypernotus or crown of the percs) when worn, sleep res when carried (your mind is a temple)

would appreciate suggestions/feedback before it gets merged. it's tested to function, not playtested yet, though I'm about to do a android clairvoyant changed lightsaber (psionic lesser magic living) run to see how it goes.

### DIFF
--- a/include/artilist.h
+++ b/include/artilist.h
@@ -1274,6 +1274,21 @@ A("Apotheosis Veil",				CRYSTAL_HELM,			(const char *)0,
 	ENLIGHTENING, (ARTI_PLUSSEV)
 	),
 
+/*
+ * does not give telepathy. if you have telepathy, turns it to active and boosts doubles range 
+ * if you don't have telepathy, nullifies mind blasts and any psychic damage
+ * makes monsters within your extrinsic telepathy range treated as if they WERE telepathic (does not include cross-level, only in your range)
+ */
+A("Enforced Mind",				HELM_OF_TELEPATHY,			(const char *)0,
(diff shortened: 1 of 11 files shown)


RikerW:
just tried this, it's mediocre at best damage-wise - you can spam deeps weapons with it i guess if more are ever added, but with active telepathy psionic oprop is still just a normal 2x damage prop so it's not anything busted, and it doesn't work vs. mindless. probably should've realized this a while ago but who cares. i avoided using good props but did through on lesser acid and lesser holy for kicks

Chris-plus-alphanumericgibberish (approved):

PR #2190, opened 2023-08-02
Finally fix dnh using .nethackrc not .dnethackrc

RikerW:
genuinely no clue why this hasn't been changed, a quick grep says this is hte only meaningful source of `.nethackrc` in the code so idk maybe its just not an issue for other ppl but this has bugged me for literal years LOL

also no clue why k2 never mentioned this but idk maybe he has a script that ignores the files.c setting

### DIFF
--- a/src/files.c
+++ b/src/files.c
@@ -1591,7 +1591,7 @@ const char *filename;
 
 const char *configfile =
 #ifdef UNIX
-			".nethackrc";
+			".dnethackrc";
 #else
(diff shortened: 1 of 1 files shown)


k21971:
It's because the Hardfought server config for each variant, it sets its own path/filename for the rc config. You would only notice the issue playing locally.

Chris-plus-alphanumericgibberish (approved):

PR #2193, opened 2023-08-07
Prevent EQ from destroying altars, and temple priests from casting it

RikerW:
EQ sources (drum, EQ spell, earth crack spell, tremors spell, etc.) never destroy altars. Sometimes this is reasonable, other times it's not. Furthermore, temple priests never cast EQ or its ilk themselves - where temple priests are defined by having `ispriest` set, so they have some assigned shrine somewhere. This should cut down on both the self-destructive priests who see nothing wrong with immediately shattering their own temple in response to a minor inconvenience, and also cut down on the amount of times the altar itself is destroyed even if it's a summon or some other angry dipshit nearby instead of the shrine priest itself.

### DIFF
--- a/src/mcastu.c
+++ b/src/mcastu.c
@@ -6535,6 +6535,14 @@ int tary;
 	if ((spellnum == EARTHQUAKE || spellnum == TREMOR || spellnum == EARTH_CRACK) && In_endgame(&u.uz))
 		return TRUE;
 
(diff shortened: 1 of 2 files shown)


RikerW:
closed as per discord discussion - eventual change will just be "altar not nuked by eq" and leave priests casting eq

PR #2195, opened 2023-08-07
Constellation pets should not be mindless

RikerW:
This makes a lot of really good pets otherwise really useless, since they don't understand the concept of things like "use a unicorn horn so you stop hitting me" or "put on the fucking armor goddamnit holy shit" consistently. I hate pet AI so that second one might not be true but the unicorn horn one sure is - try tossing it, they'll never catch it or pick it up. Mindless ki-rins or unicorns won't even use a built-in horn, if you wanted a mount.

I understand that there's probably very good lore reasons so this is definitely open to revision, and it's not like I decide how they behave anyway, but I tried using them and was VERY underwhelmed until I hacked out MT_MINDLESS (intoner and ootaos if you were curious).


RikerW:
replaced with #2208

Issue #2189, opened 2023-08-01
Casting full healing spell at self behaves inconsistently


demogorgon22:
Doing some testing it seemed to work well as a vampire ana, but as a convict human zapping it at myself it seems to do nothing? Or maybe it is just healing me a tiny bit?

RikerW:
Looks like an oversight with how full healing HP is calculated. It heals 50*healing skill, but P_ISRESTRICTED is 0, so for anybody restricted it'll always heal 0. It should probably use a more traditional formula, or instead just do min(1, skill value).

Chris-plus-alphanumericgibberish:
Right, supposed to stay open until dev goes live >_<

demogorgon22:
solution: go live with dev :D

Issue #2182, opened 2023-07-24
Several impossibles can be triggered by misbehaving in a shop


demogorgon22:
https://github.com/NetHack/NetHack/commit/e51713bccfbbd20a6a35e1618727b11dbdf40141 - this is not a panic in dnh
"splitbill: no resident shopkeeper??  Program in disorder - you should probably S)ave and reload the game.  onbill: unpaid obj not on bill? sub_one_frombill: unpaid object not on bill  Program in disorder - you should probably S)ave and reload the game."

Clearer? steps to replicate:

Enter a large shop. (need sufficient room to still be in the shop but be far enough from the shk that he will be willing to leave it once angry)
Wish for a large stack of projectiles.
Sell your projectiles and then pick them up again.
Trap shopkeeper against the door with boulders so that he cannot move anywhere else into the shop.
While still in the shop, throw the projectiles at the shopkeeper to anger him.
Move towards the back corner of the shop, away from the boulder fort and wait for the shopkeeper to leave the
shop due to your distance from him.
Throw one of the projectiles out past the boulder fort.

Issue #2180, opened 2023-07-24
Statues turned into boulders do not update the vision matrix


demogorgon22:
https://github.com/NetHack/NetHack/commit/a3341e071b0f1eb3bb2eb9c1d424df231a61d175
When polying a statue and it turns into a boulder, it is see through until it is pushed.

Issue #2179, opened 2023-07-24
Tunneling tame monsters phase on the rogue level


demogorgon22:
https://github.com/NetHack/NetHack/commit/fb6cbd37e7e73aa744a7eb73d533d34d9bd64036
They aren't supposed to dig because nobody is allowed to, but tunneling pets still get to think they can path through walls so they just kinda phase. Tested with tame umber hulk.

PR #2201, opened 2023-09-06
Misc. fixes for a large collection of minor bugs/enhancements

RikerW:
minor fixes/enhancements. mostly just simple fixes from issues, some from demo's 3.5 fixes where i just copied whatever vanilla did, others written myself. first set is mine second is 3.5 fixes. only major enhancement - warn of mon has useful messages :)))))

#2189 - Casting full healing spell at self behaves inconsistently. fh spell now heals minimum of 100 hp - this should be fine, it basically means restricted/unskilled == basic. its high enough level ur not casting it well without a caster build anyway. this prevents it from healing 0 or only a piddly 50 if u suck at healing spells (min 100 in case an artifact grants 100% success later or something to make that artifact usable on non-casters idk)
#2164 - Torches do not stack but some roles start with a stack of them. (shadow)torches now stack if (obj1->age - obj2->age)/25 rounds to 0, same as candles. this includes sunrods & magic torches but sunrods will never stack if lit to prevent what i think looks awkward there. (shadow)torches stack when lit same as candles
#2163 - Read after free in magic bag explosion code (credit demo for fix)
#2148 - Object material names can override base item name for database lookups. added arg to doxname for getting_obj_base_desc (previously static flag), which is the flag used to block recursively getting shit for finding obj arti descriptions or something. anyway its now a parameter instead and its called via wrapped xname_bland, which is doxname(obj, FALSE..., TRUE) so it just returns the basic bitch name. this should be as functional as desired for encyc lookup - you'll still get base item type, you'll still get artifact name, and at a QUICK glance that seems to be really all the encyclopedia really wants? if you want material, look that up separately with #whatis or something (since #whatis uses user input its entirely unaffected by all of this). dunno let me know if this blots out adjectives that are important
#2143 - Monsters can't recharge offhand or extra-hand weapons. holy copypaste chris lmao, anyway changed now to do that. ok, so this sounds bad - it was segfaulting occasionally when testing, but after recompiling with no changes and throwing it in and out of gdb a bunch it stopped doing that. i dont know why it was complaining if (otmp) { <do shit with otmp> } was doing shit with a zero otmp but man what the fuck. anyway it works fine i tested but hey. oh yeah this works for anything in their inventory but prefers wep -> swapwep -> anything else in no particular order, but ONLY works on the weapons they could previously charge - aka vibros, blasters, lightsabers.
#2138 - Laser beams cut trees into stones. turns out IS_STONE is for nopass things just rocks, but everything else should be rock. exempted trees and gave them a separate entry.
#2132 - Clockwork kicks can be blocked by thick skin. fixed by adding is_metallic check to the boots (in addition to kicking boots for offmat ones, and in addition to the other boots that work) and checking youagr && !Upolyd && race_if(clck). could probably check other races with big metal feet but that feels niche enough it's not really worth addressing imo
#2122 - Blade-Dancer's Dagger can be duplicated if obtained first
#2117 - Longbow of Diana weighs 1 aum. artifact weights now bypass the 0 aum -> 1 aum thing, and ONLY artifact weights.
#2107 - Binding-Amon-in-temples handling is bugged
#2099 - Incantifiers should be able to eat magic whistles. yeah let netsysfire cook bro is onto something fr
#2075 - Adding flaying oprop to an item makes ALL of its damage negated by thick skin. honestly my bad for telling demo he was hallucinating i swear my wiztests showed it worked fine. but yeah no i have no clue why this code is here, flaying doesn't add extra damage thats done in arti.c so i don't know why this change was made, probably balance, but it's stupid i think? and should be changed. the only real effect this has on the player is that flaying weapons are no long really useless, and against the player flaying weapons are _probably_ about the same? most monsters with flaying weapons don't really have absurd damage output, so I think this only really affects Bael killing you a little quicker now. but you don't want to get hit by him anyway
#2057 - Fixed drow armor does not repair when taken off. actual bug was "shadowsteel timers break when put in containers", now containers do not break and will count as "darkness" when appropriate. any other locations that aren't containers (buried, migrating) no longer break timers as well
#2056 - goat/sflame holy symbols now have obj appearance changes in inventory if you have 50+ credit (aka minor boon) or enough credit for a major boon. this is the same info u could have previously gotten by applying it, but without the -1 san per apply or whatever, dunno, it feels fair. msgs for minor/major are glistening/drooling and glinting/shining
#2054 - Weapons that go to +10 vibrate unexpectedly at 6 or higher. vibrating chance changed from at 5+ to at safelim+, so +10 weapons do this at 9+ instead. fixes them vibrating like 4 times in a row when its totally safe
#2053 - SUMMON_SPHERE monster spell is not blocked by dimension locks
#2036 - Hilite_hidden_stairs should support portals, or an option for this should be added. hilite hidden stairs now also does the same red background on magic portals. this is not complicated and does exactly what u think it does.
#1959 - Shadowsteel artifacts can be destroyed by normal shadowsteel erosion. they cap at very tenuous now
#1800 - Bullets are by default iron when a more fitting material now exists for them. i see nothing else that should be lead
#1702 - Half Fem first sac gift overrides too hard. now overrides only for roles w/o first gift, or if you're a noble (so no lordly might or anything).
#1379 - Colored branch stairs ought to be implemented for QoL improvements. now with free colored altars. so basically, they exist now. the "default" color is yellow but almost all branches are recolored to be distinct because i could - for example, mines is brown, sea is yellow, temple is red, tomb is black, etc. this includes (because reasons) outlands<->rlyeh and outlands<->dispensary but i believe that's all of the "branch" stairs in the dungeon. oh VOTD is grey/white due to being overriden by VOTD colors, but who cares, that's not a very helpful one to turn yellow. anyway, so this exists. uses S_brdnstair and S_brupstair like vanilla & other variants does, and handling should be added for all relevant things, including moving the "<= S_dnstair"s to use branch stairs instead when checking certain "end of dungeon feature" glyph lists. support has been added for DEC/IBM/UTF graphics too if barely, they render same as stairs but colored rather than as ladders. branch ladders don't exist btw so don't add those without adding the appropriate symbol i guess. altars also now get hallu colors with god's worst redraw function. i jest it's not that bad but yeah added see_altars to see_traps/_objects_monsters, which does the similar thing, but clears the glyph buffer (extra operations whoo) instead of just calling newsym because altars are all S_altar recolored, so the game needs to recompute colors, and so if other "color shifting" glyphs are added this could be problematic. shouldn't be a huge deal though since i don't expect any other things to change color per turn since they wouldn't work without a redraw anyway. hallu altars are random color, and then lnc is white/grey/black, unaligned yellow, void red (idk black is used, blue is black for a lot of ppl i think, maybe make htat magenta. up 2 chris)
#1296 - Allow training of lightsaber forms without requisite weapon skill to use them. now you are no longer blocked from viewing/enhancing those skills unless wielding a saber, or capped at your saber skill. this improves parity with 2wep skill and also the odd thing where you can be a lightsaber GOD and be shown as awful if you're not wielding one at that minute.
#1145 - Adjust how god-given minions are chosen. basically - averages depth & player level, factors in bard/healer status + bm skill for odds, and makes them loyal. odds went effectively up to 1/12 for bards/healers at xp15 w exp beast mastery, or 1/8 with max luck too.
#600 - "Warned of Foo" description is too literal. okay, so first off this is kinda an egregious level of copypaste, since this doubles up on a lot of messaging already in the same file for nero's pokedex work. the biggest issue with those is that its set up to take a monster and some booleans and shit and that's not what u want out of warnings. the function i wrote just takes a flag and a int to set which flag, set with some cheap defines (though i could probably make that more universal). it is not extended outside pager.c. it outputs the general species name for MA flags, uses symbol name from monappearances, and the rest are adjectives/phrases that felt appropriate that work grammatically with "monsters that XXXXX". it also incorporates the warning things from paimon tenebrous etc. for non-standard warning masks, so that's nice, so in the future add those to WARN_OF_MON and then here i guess
#307 - AT_WISP/AT_SHDW effect on passive attacks got lost. todo: #style for this. ill probably do that not at 3am. holy ancient issue batman this one's a post-xhity 'bug'. jesus christ.

taken from demo 3.5 stuff

#2182 - Several impossibles can be triggered by misbehaving in a shop
#2180 - Statues turned into boulders do not update the vision matrix
#2179 - Tunneling tame monsters phase on the rogue level
#2178 - is_drawbridge_wall called wrong can lead to rays not affecting certain orientation drawbridges
#2171 - Cursed destroy armor on cursed armor that affects attributes doesn't call adj_abon when spe is updated
#2160 - Death Drop weight limit uses 3.1.0 era 1/10 weight

holy shit theres been 1500 issues/prs since junethack 2020? fucked up if true

force push is because i had a thought to make dead trees that i decided to call arbitrarily petrified wood drop stone quarterstaves but i already pushed. this probably disagrees with other dead tree messages but hey who's laser cutting a tree anyway? it's neat

TODO (not necessarily before this is merged):
- take another look at how god_gives_benefit really shakes out. it's really kinda some stupid and arbitrary odds - would be nice if that was changed to be more logical
- #style option to use only passive attacks
- is_metallic or something similar for big boys with hefty kicks (re clockwork kicks).


RikerW:
this PR/branch is done now, unless I come up with something important to add - further changes to mattrib scores will likely be done elsewhere since i don't expect to need to adjust any of the functions that i've written for knight styles etc.

Chris-plus-alphanumericgibberish (approved):

PR #2214, opened 2023-10-22
fix binder astral rarely crashing constantly

RikerW:
due to a never-interacted-with off-by-one in gnames, the value of MAX_GOD was actually being used as NUM_GODS, aka 1 higher than what it should.

this fixes that - MAX_GOD is now the last one (currently GOD_NYAR). this is the same situation as nrofartifacts (ART_TRAPPINGS_OF_THE_GRAVE & NROFARTIFACTS are both 324)

the various malloc calls are fine, previously they actually had one god struct's worth of extra space, so the only thing that needed adjustment is the save/restore code to change <MAX_GOD to <MAX_GOD+1

the actual bug here was that binder astral could rarely spawn (1/106 per applicable monster) minions/priests/whatever of null terminator gods due to maxrolling a d106. this was reported by klepto, poor fucker had it crash apparently 20+ times due to the walking segfault going on before finally successfully finishing his ascension

### DIFF
--- a/src/pray.c
+++ b/src/pray.c
@@ -4711,7 +4711,7 @@ int fd;
 	godlist = malloc(sizeof(struct god) * (MAX_GOD+1));
 	mread(fd, (genericptr_t) godlist, sizeof(struct god) * (MAX_GOD+1));
 	/* fix name pointers -- assumes that god names do NOT get changed during the game */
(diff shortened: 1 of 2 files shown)


RikerW:
image

lol

Chris-plus-alphanumericgibberish (approved):

PR #2233, opened 2023-12-10
no more fiend herds

RikerW:
removes the possibility for the chaos1 statues to animate - they're historic now


RikerW:
shoutout thebobness, got immediately jumped by tiamat, fiend of wind & friends (doppelgangers), within like 10 turns of entering chaos1 lol

Chris-plus-alphanumericgibberish (approved):

PR #2245, opened 2024-01-10
More noisy fixes

RikerW:
Also a minor fix about vegan messages


Chris-plus-alphanumericgibberish:
Added and adapted via cherry pick on local

PR #2258, opened 2024-02-20
Allow merging stacks with unknown but identical traits

RikerW:
This only changes behavior for 3 fields - known, bknown (BUC), rknown (rustproof). stolen is already handled and already merges or auto-IDs state, and dknown (color) has not been changed. i don't know what that would affect, but i think the idea is that dknown is set if you can see it, so if you don't have dknown you don't know ANYTHING. this is also keeping in line with vanilla/etc. re dknown

This does potentially affect behavior for unknown oprops, since I believe those use ->known. I consider this intentional.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -187,6 +187,11 @@ struct obj **potmp, **pobj;
 			    / (otmp->quan + obj->quan);
 
 		otmp->quan += obj->quan;
+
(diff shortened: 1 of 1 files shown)


Chris-plus-alphanumericgibberish:
I put in the rknown and dknown change, but I'm leaving known and bknown as they currently are.

PR #2260, opened 2024-02-21
Clockwork Monk inspired fixes

RikerW:
Local clockwork monk, descendent silver starlight + yog, unearthed a bunch of weird bugs + enhancements.

- thrown weapons get dbon as they should
- clockworks/incantifiers have their satiation limit set to 3/4 (and their "hard to force down" limit at 7/8), since they incentivize staying close to hungermax more than normal
- clockwork heat is now visible in both ctrl-x and in enlightenment (exact temp in wizmode enlightenment), since it's weird it wasn't before.
- spirit duration being different from timeout is now shown in ctrl-x, since this is noteable if you're yogging all over the place and kept increasing those binding durations. this does not apply to the view when you go to engrave a seal, only the ctrl-x.
- monks get an extra 3 squares radius (so 6) from EOTO xray because 3 squares of xray kinda sucks in comparison to like, nenya, orthos, scent, idk a lot of things. now they have more.
- curses display fixed for bonus str. i dont know if this made it to your local yet.
- create ammo invokes will ID the ammo created up to what you know of the original artifact. so if you have a fully ID'd silver starlight it makes fully ID'd shuriken, for example.

originally this was on top of my other PR but i dont believe it depends on it. it doesn't seem to when rebased at least

EDIT - next commit

- fixes enl bad wording from previous commit, see scienceball's comment
- don't hide upgrades behind beginner status - this very rarely makes a difference, you'll almost always get out of beginner zone before this. for reference - assuming you've gained score only via kills, that's just after xp6 - 500 xp worth of kills total. Scrap maw & phase engine are now 'visible', since it's highly unlikely that a 'beginner' would throw an upgrade kit into oil stove and have another one already there, again before xp7.
- fix an infinite-ish loop where melting down and lifesaving would just. keep melting down every turn until you chuffed water.... except that adding more water wouldn't immediately fix it, since you lose 1 temp/turn. lifesaving now sets you just under melted, so you're still taking damage, but you won't get any hotter and die for now. also adds a message about what killed you lol not just "you are melting" for 30 dmg, 30 dmg, 30 dmg, whoops theres literally 500 hp in one go

### DIFF
--- a/src/enlighten.c
+++ b/src/enlighten.c
@@ -978,11 +978,17 @@ boolean dumping;
 			enl_msg("Your internal boiler ", "is", "was", " running rather hot");
 		else if (u.utemp < MELTING)
 			enl_msg("Your internal boiler ", "is", "was", " burning hot");
-		else if (u.utemp < MELTED)
-			enl_msg(Fire_resistance ? "Your thermal sinks " : "Your boiler ", "is", "was", Fire_resistance ? " nearly at capacity" : " melting to slag");
-		else 
-			enl_msg(Fire_resistance ? "Your thermal sinks " : "Your internal boiler ", "is", "was", Fire_resistance ? 
-				" are well beyond capacity, but miraculously intact" : " nothing but molten bronze");
+		else if (u.utemp < MELTED){
+			if (Fire_resistance)
+				enl_msg("Your thermal sinks ", "are", "were", " nearly at capacity");
+			else
+				enl_msg("Your intenal boiler ", "is", "was", " melting to slag");
+		} else {
+			if (Fire_resistance)
+				enl_msg("Your thermal sinks ", "are", "were", " well beyond capacity, but miraculously intact");
+			else
+				enl_msg("Your intenal boiler ", "is", "was", " nothing but molten bronze");
+		}
 		if (wizard) {
 			Sprintf(buf, " %d", u.utemp);
 			enl_msg("Your boiler temperature ", "is", "was", buf);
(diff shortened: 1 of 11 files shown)


src/enlighten.c
```diff
@@ -970,6 +970,23 @@ boolean dumping;
...
+			enl_msg("Your internal boiler ", "is", "was", " running mildly warm");
+		else if (u.utemp < BURNING_HOT)
+			enl_msg("Your internal boiler ", "is", "was", " running rather hot");
+		else if (u.utemp < MELTING)
+			enl_msg("Your internal boiler ", "is", "was", " burning hot");
+		else if (u.utemp < MELTED)
+			enl_msg(Fire_resistance ? "Your thermal sinks " : "Your boiler ", "is", "was", Fire_resistance ? " nearly at capacity" : " melting to slag");
+		else 
+			enl_msg(Fire_resistance ? "Your thermal sinks " : "Your internal boiler ", "is", "was", Fire_resistance ? 
+				" are well beyond capacity, but miraculously intact" : " nothing but molten bronze");
```
ScienceBall: If you have fire resistance, then I think you would get a message like "Your thermal sinks **is are** well beyond capacity, but miraculously intact." The "nearly to capacity" case also has a similar issue. The cause of the issues is that "Your thermal sinks" is plural and "Your (internal) boiler" is singular. I'm not sure how to work around to make something that's grammatically correct in every case.

Chris-plus-alphanumericgibberish (approved):

PR #2269, opened 2024-03-23
Adjust plane of water to use PUDDLE terrain for the inside of bubbles

RikerW:
Also change the behavior of rust damage via puddle to NOT message when you step in them with non-rustable boots on, since it's really message spammy in certain areas (like the new plane of water bubble puddles).

Fixes up a bunch of other now-unnecessary `is_pool` checks, since `PUDDLE` is a different terrain type, and only is checked if you're calling `is_pool(..., TRUE)`. A bunch of these were things that assume real pools are water you fall into or similar, or 3d water rules, which obviously don't matter on the water plane.

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -223,7 +223,7 @@
  * "cover" any objects or traps that might be there.
  */
 #define covers_objects(xx,yy)						      \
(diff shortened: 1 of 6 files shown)


Chris-plus-alphanumericgibberish:
The plane of water should be pools, not puddles.

Chris-plus-alphanumericgibberish:
I'm gonna grab the "don't message spam for iron boots" change though.

Issue #2276, opened 2024-04-17
Non-wizmode #vanquished


NetSysFire:
3.7 already has that (NetHack/NetHack@f6b3b968e7a274480fbb9d1783930ba8cbbfb870). I am asking for extinctionist purposes so I'd even be happy to just know which creatures are extinct already.

Making this feature request formal so it does not get lost in the ##nethack-variants log.

The main concern I got so far is that you can use it to find out what invisible monster you just killed but that is pretty minor in my opinion.
In that case you could only update the list every 500+1d50 turns or so, but maybe this concern is small enough that `#vanquished` can be there unfiltered.

PR #2277, opened 2024-04-17
Show extinct monsters in ctrl-x/attributes geno listing

RikerW:
tested via going to axus as an invulnerable dude and killing 200 billion monotons and it worked


RikerW:
addresses but does not solve #2276

Chris-plus-alphanumericgibberish (approved):

PR #2296, opened 2024-09-26
Rebalance Amnesia, removing map effects and draining maximium skills instead

RikerW:
Amnesia effects now lose you x% of your total skill points, instead of your map knowledge. This does NOT refund your spent points if it takes them away from a skill. This is intended to make amnesia a much more severe threat in terms of how ruinous it is, appropriate for how few sources of amnesia there are. This may necessitate certain changes to amnesia-causing effects - in particular, I see AD_LETHE damage being 25% as potentially too messed up for how easy it might be to apply it, though I'm not sure (without checking - that's just leviathan and lethe E?). If necessary, the percentages for appropriate effects should be turned down instead of modifying the function to apply the damage - that way they're real percentages.


Chris-plus-alphanumericgibberish:
I think it needs some sort of recovery option. Hard to say what exactly though. Effigies are perhaps (often) rare enough to be considered.

PR #2319, opened 2025-04-03
Remove the bad bugfix for ghost symbols setting

RikerW:
The ghost symbol will now be mismatched if your symset changes the floor. This fixes the bug where ghost rendering badly broke if you didn't manually set a ghost symbol in your rcfile, and made certain levels borderline unplayable if you didn't know the right fix.


Chris-plus-alphanumericgibberish:
I'm still hoping to figure out what the conflicting code is, but I'll merge this if I can't

Chris-plus-alphanumericgibberish:
The conflicting code was found, so there is no longer a need to revert the bad bugfix!

PR #2330, opened 2025-05-08
add option to select no inherit artifact when descendant is set

baoliejuan:
add option to select no inherit artifact when descendant is set so don't need to constantly edit .dnethackrc to disable and enable it


baoliejuan:
just found out this is already implemented when you don't specific descendant in rcfile

PR #2354, opened 2025-12-25
treat G_O_O as 2x, not 3x, for mortal blade

RikerW:


Chris-plus-alphanumericgibberish:
Merged via cherry pick

PR #2360, opened 2026-01-12
#NoElfRings2026

RikerW:
[Acid potion shows up in poisoning](https://github.com/Chris-plus-alphanumericgibberish/dNAO/commit/427071e2cc1b8fe3ab095b6fa283ec5cbfdcf39e)

That way you can name it immediately. It's the only non-ambiguous potion (because I assumed when I changed it that acid would always corrode. no, dipshit, materials exist now (?)).

[Ban self-willed race-favoring artis from gifting](https://github.com/Chris-plus-alphanumericgibberish/dNAO/commit/29ebeb60b03a92e86d80ac02bf3bf988e84374f4)

They blast you if you're not that race :((((

[Anguirel counts as obsidian too](https://github.com/Chris-plus-alphanumericgibberish/dNAO/pull/2360/commits/558dc01d6bcd7921c11b74069e41032e323e4a08)

Fixes dice override (by moving it, really) so that it's actually comparable, damage-wise, to a normal droven greatsword (rather than being _noticeably_ worse, due to the more consistent damage not interacting well with GWF).

Add obj_is_material for Anguirel & obsidian specifically, which is only used in the Rising Cut move case for Drow Kensei. All other checks that I see are related to object destruction, which is NOT what obj_is_material should probably be linked to.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -4879,7 +4879,7 @@ use_dissection_kit(struct obj *obj)
 		// pline("That's too insubstantial to dissect.");
 		// return;
 	// }
-	splitobj(otmp, otmp->quan - 1L);
+	if (otmp->quan > 1) splitobj(otmp, otmp->quan - 1L);
 	consume_obj_charge(obj, TRUE);
 
 	//San check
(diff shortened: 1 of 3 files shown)


RikerW:
relevant bits implemented

PR #2361, opened 2026-01-20
Material dmod changes based on attack mask

RikerW:
Previously, only base object type entry was considered. This was inaccurate for a variety of edge cases (launchers that work in melee, artifacts with additional or removed damage masks, fighting forms & expert trait forms that enforce specific damage masks, etc.).

Outside of material dmod checks, `is_stabbing` and `is_bludgeon` were only used for Smaug & erosion respectively. `is_slashing` and `is_blasting` were entirely unused. As a result, the macros have been removed and replaced with `attack_mask` calls.

Does not break save compatibility (tested to make double sure).


RikerW:
cherrypicked

PR #2364, opened 2026-01-29
Only fix petrification when actually getting petrified

Shrigis1:


Chris-plus-alphanumericgibberish:
I have a different fix on devel (always fix petrification and avoid duplicate messages by having an empty string for gaining the intrinsic).

Shrigis1:
Wouldn't that print the wrong message if you gain it while hallucinating? `What a pity - you just ruined a future piece of art!`

PR #2366, opened 2026-01-31
Allow wishing for specific Sky Reflected weapon type

baoliejuan:
instead of

> Born of level 0 of The Dungeons of Doom."  You cannot wish for an existing artifact.


RikerW:
i don't think this is intended, and i'd actually hazard this is probably a bad change for build diversity. to be kept in check it'd likely need a similar whitelist to fire brand/frost brand, and being able to get your choice of traits/weapon specific abilities on an amalgamated skies on any run (not just gz kensei) is probably a bad idea

PR #2368, opened 2026-01-31
Esscoo shouldn't pick up bolted magic chests

baoliejuan:


Shrigis1:
Moving magic chests after bolting them can actually be pretty useful, so in a way this is a nerf. Maybe it should unbolt them first instead?

Chris-plus-alphanumericgibberish (approved):

PR #2380, opened 2026-03-25
Ruinous descent should favor drow

Shrigis1:
The drow race flag got removed from it in commit `https://github.com/Chris-plus-alphanumericgibberish/dNAO/commit/ecc370191328d2f06b8677ade2e58b59cfd54c36` 
alongside a bunch of drow quest artifacts. 
This made it unusable without either permaconverting or wearing a HoOA. 
Looking through the rest of the crowning gifts for non-starting gods, it appears that the only artifacts that share this property are The Red Cords of Ilmater (not sure if that's intentional) and gnome ranger crowning artifacts, which I assume is done deliberately to inconvenience the player, same as everything else relating to the gnome race.
Either way liecleaver, which is also a drow noble crowning gift, but lawful instead of chaotic, does favor drow. Plus Ghaunadaur, who gives you the artifact, has hedrow minions

### DIFF
--- a/include/artilist.h
+++ b/include/artilist.h
@@ -2383,7 +2383,7 @@ A("Liecleaver",						DROVEN_CROSSBOW,	(const char *)0,
 /* Hedrow noble crowning gift, Chaotic */
 A("The Ruinous Descent of Stars",	MORNING_STAR,		"silver-spiked %s",
 	8000L, METAL, MZ_DEFAULT, WT_DEFAULT,
-	A_CHAOTIC, NON_PM, NON_PM, TIER_A, (ARTG_NOGEN|ARTG_NOWISH|ARTG_MAJOR|ARTG_FXALGN),
+	A_CHAOTIC, NON_PM, PM_DROW, TIER_A, (ARTG_NOGEN|ARTG_NOWISH|ARTG_MAJOR|ARTG_FXALGN),
 	NO_MONS(),
 	ATTK(AD_PHYS, 1, 0), (ARTA_SILVER),
 	PROPS(), NOFLAG,


Shrigis1:
Actually upon further inspection The Red Cords of Ilmater is usable normally, it just gets the role set at the start of the run. So ruinous descent is definitely broken

RikerW:
i believe this is intentional? similar to gnome ones, it's meant to be "this is strictly aligned and not actually race favoring", so you do actually have to commit to being that align to use it. alt crowning gods are designed to be converted to, not crowned on and forgotten about, despite how easy that is for most roles

Shrigis1:
Maybe I'm reading the code wrong, but from what I'm seeing in artilist.h literally every other crowning gift besides the gnome ones is either:
1) aligned to the role (sunbeam, dark claymore, all 5 archeologist gifts, even ibite arm)
2) aligned to the race (all 3 elf gifts) 
3) lacks the ARTG_MAJOR flag (sickle moon)
4) goes out of its way to be used without converting (red cords)

I believe the gnome artifacts are literally a joke at the expense of the player and are deliberately inconvenient to use

RDoS is also the only artifact which stopped being usable normally after that commit, since going traitor is a free conversion and a part of the quest, while becoming chaotic as a drow noble/healer is something far out of the way. If this were intentional, I expect there to be a comment in the code or at least a mention of this in the commit text

From a gameplay perspective crowning under an unusual god already has an opportunity cost since the player has to miss out on their usual crowning gift. 

Ghaunadaur, who gives you the RDoS, has more drow minions than Ver'tas, while Ver'tas's artifact (Liecleaver) is aligned to drow as expected

Please correct me if I'm wrong or provide examples of artifacts that don't follow this pattern (i. e. the player has to remain the artifact's alignment after crowning)

Chris-plus-alphanumericgibberish:
I think this is emergent behavior from Ghaunadaur being treated as a non-drow-specific god.

Almost all pantheons are composed of gods that belong to the same culture the PC belongs to. Ghaunadaur, Kurtulmak, and Urdlen might be the only evil gods that belong to a different culture than the PC's.

I'm very unsure what the correct behavior around crowning is. It actually seems mildly bug-y that the PC can crown under an "incorrect" god using a HoOA. But at any rate it doesn't seem clearly wrong that Ghaun requires greater commitment than many gods.

Shrigis1:
To me it definitely feels buggy that when crowning with a helm of opposite alignment there is AFAIK no way for the player to know if they will be able to use the artifact they get or not without source diving. There is already a precedent for crowning artifacts to always be useable for all but 3 (and 2 of them are for gnomes, a joke race). 
Artifact blasting already serves as a way of punishing the player for being the wrong alignment. 
If we are actually going the route of gods requiring different amounts of commitment, I feel a better solution would be harsher blasts per god rather than just telling the player "NO".

I still believe Ghaunadaur should count as a fan of drow, since 
1) He summons drow to fight for him as minions
2) data.base states that `He is worshiped mainly by disaffected drow` 

But if that isn't an option, then how about making ruinous descent usable through the role flag for (drow) nobles/healers (exactly like the rod cords do it) but for example give the artifact blast a 10% chance to start delayed sliming (cuz Ghaunadaur is also a slime guy). That would be a lot more fun and unique than just saying "nope, sorry, go permaconvert"

Chris-plus-alphanumericgibberish:
c6f0450be9d26412bceb7a385c13cd4989b0c0a7 Adjust RDoS to match your role and alignment if Ghanadaur is your chaotic god.

